PR #30, opened 2022-07-14
[ENG-1165]: Towards an abstraction around Vehicle/Fleet States

dirkschumacher:
This PR prepares the types needed to implement some remaining router options. It introduces some new abstractions around the vehicle/fleet states. More in the ticket.


larsbeck (dismissed):

route/option.go
```diff
@@ -232,6 +236,103 @@ func ServiceGroups(serviceGroups []ServiceGroup) Option {
...
+
+// Sorter sets the given custom vehicle sorter. The vehicle sorter lets you
+// define a function which returns the vehicle indices in a specific order. The
+// underlying engine will try to assign the locations to each vehicle in that
+// returned order.
+func Sorter(sorter func(
+	p FleetPlan,
+	locations model.Domain,
+	vehicles model.Domain,
+) []int,
```
sebastian-quintero:
```suggestion
func Sorter(
	sorter func(p FleetPlan, locations model.Domain, vehicles model.Domain) []int,
```

I think this looks nicer and:
1. Aligns with `Filter` in that the argument starts in a new line.
2. The linter doesn't complain about the line length, so it is good to have all the `func` args in one line.

route/option.go
```diff
@@ -232,6 +236,103 @@ func ServiceGroups(serviceGroups []ServiceGroup) Option {
 	return serviceGroupsFunc(serviceGroups)
 }
 
+// Selector sets the given custom location selector. The location selector lets
+// you define a function which selects the location that will be inserted next
```
sebastian-quintero:
```suggestion
// you define a function which selects the locations that will be inserted next
```

From ignorance I would expect that `model.Domain` holds multiple locations? If only one location will be inserted should it be an `int`? We should check to make sure 🙏🏻 .

route/router.go
```diff
@@ -39,6 +40,23 @@ type Router interface {
 	Plan() store.Variable[Plan]
 }
 
+// FleetPlan is an (incomplete) Plan that operates on the internal
+// solver data structures. Certain router options that customize solver
+// internals have to work with this data structure.
+type FleetPlan struct {
```
sebastian-quintero: Comparing this and `FleetVehicle` to `Plan` and `PlannedVehicle` one can see the similarities. Would it be possible to simply use `Plan` and `PlannedVehicle` and maybe adding information such as `Unplanned`? Would make it simpler.

sebastian-quintero (commented):
I would like to get together to discuss some potential changes.

route/router.go
dirkschumacher: I was first experimenting with that idea but settled with two different structs because the data structures used internally (ints mostly) and the data structures of a user facing result (Stops, Vehicles) are fundamentally different. In addition it is quite expensive at the moment to convert the internal data structures to an information richer format.

route/option.go
dirkschumacher: Yeah was thinking the same, it mimics the current API though. Maybe the reasoning was to also be able to return an empty domain? @larsbeck @nilsbeck @merschformann what do you think?

larsbeck (dismissed):

route/router.go
```diff
@@ -39,6 +40,41 @@ type Router interface {
...
+//  - Vehicles(): a slice of vehicles part of this partial plan.
+//  - Value(): the value of this plan.
+type PartialPlan interface {
+	Unplanned() model.Domain
+	Unassigned() model.Domain
+	Vehicles() []PartialVehicle
+	Value() int
+}
+
+// PartialVehicle represents a Vehicle that operations on the internal solver
```
sebastian-quintero:
```suggestion
// PartialVehicle represents a Vehicle that operates on the internal solver
```

route/router.go
```diff
@@ -39,6 +40,41 @@ type Router interface {
...
+// data structures. Certain router options that customize solver internals have
+// to work with this data structure. It gives access to the following
+// information:
+//  - Route(): the route of the vehicle represented by a sequence of stop
+//	  indexes. The first and last index is always the start and stop of the
+//	  vehicle.
```
sebastian-quintero:
```suggestion
//  - Route(): the route of the vehicle represented by a sequence of stop
//	  indexes. The first and last indexes are always the starting and ending locations of the
//	  vehicle, respectively.
```

Just a suggestion.

route/router.go
```diff
@@ -39,6 +40,41 @@ type Router interface {
...
+// internals have to work with this data structure.
+// It gives access to the following information:
+//  - Unplanned(): an Integer Domain with not yet assigned or unassigned
+//	  stops indexes.
+//  - Unassigned(): an Integer Domain with unassigned stop indexes.
+//	  These are stops explicitly excluded from being served by a vehicle.
+//  - Vehicles(): a slice of vehicles part of this partial plan.
+//  - Value(): the value of this plan.
+type PartialPlan interface {
+	Unplanned() model.Domain
```
sebastian-quintero: We tend to organize the interface methods in alphabetical order. We try, at least.

route/router.go
```diff
@@ -39,6 +40,41 @@ type Router interface {
...
+// information:
+//  - Route(): the route of the vehicle represented by a sequence of stop
+//	  indexes. The first and last index is always the start and stop of the
+//	  vehicle.
+//  - ID(): the string vehicle ID.
+//  - Value(): the value of vehicle.
+//  - Updater(): in case a custom VehicleUpdater is used it can be accessed
+//	  using this function. nil in case no VehicleUpdater was used.
+type PartialVehicle interface {
+	Route() []int
```
sebastian-quintero: Same here, may we organize in alphabetical order?

route/router.go
```diff
@@ -39,6 +40,41 @@ type Router interface {
...
+//  - Unplanned(): an Integer Domain with not yet assigned or unassigned
+//	  stops indexes.
+//  - Unassigned(): an Integer Domain with unassigned stop indexes.
+//	  These are stops explicitly excluded from being served by a vehicle.
+//  - Vehicles(): a slice of vehicles part of this partial plan.
+//  - Value(): the value of this plan.
+type PartialPlan interface {
+	Unplanned() model.Domain
+	Unassigned() model.Domain
+	Vehicles() []PartialVehicle
```
sebastian-quintero:
```suggestion
	// A slice of vehicles part of this partial plan.
	Vehicles() []PartialVehicle
```
I'm going to leave this comment on this function but it should be applied to all functions. The description for the function should be placed as a normal godoc, so that intellisense will show you the documentation when attempting to summon the function. Otherwise, when you hover over the function, no documentation is shown. In general, interface methods are documented on the method definition itself. After moving the documentation of this method here, we can delete it from the general `PartialPlan` godoc.

route/option.go
```diff
@@ -232,6 +236,228 @@ func ServiceGroups(serviceGroups []ServiceGroup) Option {
...
+// arguments, the constraint to be applied and a list of vehicles, indexed by
+// ID, to which the constraint shall be applied.
+func Constraint(constraint VehicleConstraint, ids []string) Option {
+	connect()
+	return constraintFunc(constraint, ids)
+}
+
+// VehicleConstraint defines an interface that needs to be implemented when
+// creating a custom vehicle constraint.
+type VehicleConstraint interface {
```
sebastian-quintero: Stupid and irrelevant but if you wish to keep consistency, this type should be declared above the function that uses it: `Constraint`.

route/option.go
```diff
@@ -232,6 +236,228 @@ func ServiceGroups(serviceGroups []ServiceGroup) Option {
...
+	Violated(PartialVehicle) (VehicleConstraint, bool)
```
sebastian-quintero: For the sake of clarity, can we add a godoc here that states something like `Violated returns true if the vehicle with the current route is not operationally valid`  or something to that extent?

sebastian-quintero: And also something that points to the reasoning why you need to return the constraint. Like, when would you alter the constraint...

route/option.go
```diff
@@ -232,6 +236,228 @@ func ServiceGroups(serviceGroups []ServiceGroup) Option {
...
+		locations model.Domain,
+		routes [][]int,
+	) model.Domain,
+) Option {
+	connect()
+	return filterWithRouteFunc(filter)
+}
+
+// Sorter sets the given custom vehicle sorter. The vehicle sorter lets you
+// define a function which returns the vehicle indices in a specific order. The
```
sebastian-quintero, line 315: Here the word is `indices` and elsewhere it is `indexes` (in `router.go`). Can we settle for one style and apply it throughout?

route/option.go
```diff
@@ -232,6 +236,228 @@ func ServiceGroups(serviceGroups []ServiceGroup) Option {
...
+fleet level is updated.
+
+To customize the value function that will be optimized, the third parameter in
+either of the interfaces must be true. If the last parameter is false, the
+default value is used and it corresponds to the configured measure.
+
+To achieve efficient customizations, always try to update the components of the
+state that changed.
+
+Example - VehicleUpdater Sample Implementation
```
sebastian-quintero: Because there is a subsequent ticket to implement examples for each of these new options, [ENG-1282](https://linear.app/nextmv/issue/ENG-1282/add-router-examples-as-tests-to-sdk), I would not have these examples are part of the main godoc but rather rely on the testable example to document all of this. That means that the godoc for the `Update` option should stop at line 290 and the rest should be incorporated in the subsequent `func ExampleUpdate_vehicleUpdater` and `func ExampleUpdate_planUpdater` functions. This is just an opinion though.

route/option.go
```diff
@@ -232,6 +236,228 @@ func ServiceGroups(serviceGroups []ServiceGroup) Option {
...
+// PartialVehicles as an input and returns three values, a PlanUpdater with
+// potentially updated bookkeeping data, a new solution value and a bool value
+// to indicate wether the vehicle's solution value received an update.
+// See the documentation of Update() for example usage.
+type PlanUpdater interface {
+	Update(PartialPlan, []PartialVehicle) (PlanUpdater, int, bool)
+}
+
+/*
+Update sets the collection of functions that are called when transitioning from
```
sebastian-quintero, line 269: It would be kind of nice to allude to a `store` here, as opposed to `state`.

route/option.go
```diff
@@ -232,6 +236,228 @@ func ServiceGroups(serviceGroups []ServiceGroup) Option {
...
+}
+
+// VehicleUpdater defines an interface that is used to override the vehicle's
+// default value function. The Update function takes a PartialVehicle as an
+// input and returns three values, a VehicleUpdater with potentially updated
+// bookkeeping data, a new solution value and a bool value to indicate wether
+// the vehicle's solution value received an update.
+// See the documentation of Update() for example usage.
+type VehicleUpdater interface {
+	Update(PartialVehicle) (VehicleUpdater, int, bool)
```
sebastian-quintero, line 252: In this case I think it is ok that the godoc is only on top if the interface because there is a single method.

route/option.go
```diff
@@ -232,6 +236,228 @@ func ServiceGroups(serviceGroups []ServiceGroup) Option {
...
+// the lowest index will be inserted next.
+func Selector(selector func(PartialPlan) model.Domain) Option {
+	connect()
+	return selectorFunc(selector)
+}
+
+// VehicleUpdater defines an interface that is used to override the vehicle's
+// default value function. The Update function takes a PartialVehicle as an
+// input and returns three values, a VehicleUpdater with potentially updated
+// bookkeeping data, a new solution value and a bool value to indicate wether
```
sebastian-quintero, line 248: Here we have `bookkeeping` and elsewhere we have `book-keeping`. Can we settle for one style and apply it throughout?

route/option.go
```diff
@@ -232,6 +236,228 @@ func ServiceGroups(serviceGroups []ServiceGroup) Option {
...
+// the vehicle's solution value received an update.
+// See the documentation of Update() for example usage.
+type VehicleUpdater interface {
+	Update(PartialVehicle) (VehicleUpdater, int, bool)
+}
+
+// PlanUpdater defines an interface that is used to override router's default
```
sebastian-quintero:
```suggestion
// PlanUpdater defines an interface that is used to override the router's default
```

route/option.go
```diff
@@ -232,6 +236,228 @@ func ServiceGroups(serviceGroups []ServiceGroup) Option {
...
+// value function. The Update function takes a Plan and a slice of
+// PartialVehicles as an input and returns three values, a PlanUpdater with
+// potentially updated bookkeeping data, a new solution value and a bool value
+// to indicate wether the vehicle's solution value received an update.
+// See the documentation of Update() for example usage.
+type PlanUpdater interface {
```
sebastian-quintero, line 264: I think it would be good to add _why_ the `Update` function receives a slice of vehicles. I believe in `code` we say something regarding that those are the vehicles that changed or something like that.

sebastian-quintero (dismissed):
Requesting changes for some consistency in the style and otherwise just leaving suggestions. Amazing 🌮 .

route/option.go
dirkschumacher: Replaced by store for now.

route/option.go
```diff
@@ -229,6 +233,127 @@ func ServiceGroups(serviceGroups []ServiceGroup) Option {
...
+	Update(PartialVehicle) (VehicleUpdater, int, bool)
+}
+
+// PlanUpdater defines an interface that is used to override the router's
+// default value function. The Update function takes a Plan and a slice of
+// PartialVehicles as an input and returns three values, a PlanUpdater with
+// potentially updated bookkeeping data, a new solution value and a bool value
+// to indicate wether the vehicle's solution value received an update. The
+// second parameter is a slice of PartialVehicles which may have been updated.
+// All vehicles not part of that slice have not definitely not changed. This
```
sebastian-quintero: Have not definitely not changed? I think this is what you wanted?
```suggestion
// All vehicles not part of that slice have definitely not changed. This
```

sebastian-quintero (changes requested):
One last clarification needed.

route/option.go
dirkschumacher: good catch

route/doc.go
```diff
@@ -184,10 +184,10 @@ point into the higher dimension by appending 0s.
 
 Indexed Measures
 
-Models that do not require points operate on indices. These indices may or
```
larsbeck: I believe indices is correct. Maybe a native speaker has a different opinion? [See here](https://english.stackexchange.com/questions/61080/the-plural-of-index-indexes-or-indices)

larsbeck (dismissed):

route/doc.go
dirkschumacher: Let's go for indices then

larsbeck (approved):

sebastian-quintero (approved):

PR #31, opened 2022-07-14
Scan working directory for plugins (first)

larsbeck:
Currently SDK looks for plugins in the directory configured via NEXTMV_LIBRARY_PATH and if that is not set, then it looks in the user's home directory under .nextmv/lib . We also want it to look in the current working directory.

### DIFF
--- a/plugin/connect.go
+++ b/plugin/connect.go
@@ -9,7 +9,6 @@ import (
 	"plugin"
 	"reflect"
 	"runtime"
-	"sync"
 
 	"github.com/nextmv-io/sdk"
 )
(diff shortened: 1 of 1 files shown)


plugin/connect.go
```diff
@@ -47,38 +56,12 @@ func Connect[T any](slug string, name string, target *T) {
...
-var mtx sync.Mutex
-
-func loadPlugin(slug, path string) (*plugin.Plugin, error) {
-	// Only load the plugin once. Then reuse the plugin pointer.
-	if p, ok := loaded[slug]; ok {
-		return p, nil
-	}
-
-	mtx.Lock()
-	defer mtx.Unlock()
```
merschformann, line 61: FYI for anyone else reading this: `plugin.Open(path)` does not need to be protected from concurrent access, since it does that already by itself (see [here](https://pkg.go.dev/plugin#Open) - "It is safe for concurrent use by multiple goroutines."). This simplifies the code quite a bit.

plugin/connect.go
```diff
@@ -19,18 +18,28 @@ import (
...
+	pluginPath := ""
+	for _, path := range paths {
+		if _, err = os.Stat(path); !errors.Is(err, os.ErrNotExist) {
+			pluginPath = path
+			break
+		}
+	}
+	if pluginPath == "" {
+		fmt.Fprintf(os.Stderr,
+			"could not find plugin %q in any of the paths %q\n\n", slug, paths)
```
merschformann, line 36: What does the `%q` do here? It's used for the string too, right? There are other places where it's used too.

plugin/connect.go
```diff
@@ -19,18 +18,28 @@ import (
...
 		os.Exit(1)
 	}
 
-	p, err := loadPlugin(slug, path)
+	p, err := plugin.Open(pluginPath)
```
merschformann, line 40: So we essentially attempt to load the first file we find / that exists, right? I like it. :+1:

merschformann (approved):
Nice work! Just had a simple question, nothing blocking. :rocket:

plugin/connect.go
larsbeck: To double-quote strings as in Go source, use %q.

I guess we can change it to %s. Let me do that.

larsbeck: Actually, looking at both outputs, I like the %q better :-)

PR #34, opened 2022-08-01
ENG-1300 ENG-1334 Simplify store generation

sebastian-quintero:
# Description

Simplify the generation of new stores by changing the signature of the `Store.Generate` func:

```go
Generate(func(Store) Generator) Store
```

A `Generator` may be obtained by using `Eager` or `Lazy`:

```go
func Eager(s ...Store) Generator
func Lazy(c Condition, f func(Store) Store) Generator
```

All other interfaces and unnecessary functions were discarded. In addition, all the examples were updated and the sudoku example was deleted pending revision. Lastly, another method was added to the main `Store` interface: `Validate`. It receives a `store.Condition` to establish operational validity.

## Changes

- Deleted `store.If`, `store.Scope`.
- Deleted `store.Action` and changed `store.Generator` to be `any`.
- Added `store.Eager`, `store.Lazy` and changed the signature of `Store.Generate`.
- Added `.Validate` to the `Store` interface.


davidrijsman (changes requested):
I remember discussing this before, Eager and Lazy allow the user to do the same thing in two ways. One will be less efficient than the other (in terms of memory). Why would we allow the user to do the same in two ways and one of them being less efficient. I believe we concluded the last time lazy is enough to expose. Is there a reason we need both ways?

sebastian-quintero:
@davidrijsman And I met to discuss that `Eager` and `Lazy` are both there to make it more ergonomic for the user.

larsbeck (approved):

davidrijsman (approved):

PR #35, opened 2022-08-02
Bump version to v0.16.0-dev.0-5

sebastian-quintero:


larsbeck (approved):

davidrijsman (approved):

VERSION
```diff
@@ -1 +1 @@
-v0.16.0-dev.0-4
\ No newline at end of file
+v0.16.0-dev.0-5
```
merschformann, line 1: No new line at end of file? :thinking:

merschformann (approved):
:rocket:

VERSION
sebastian-quintero: No, because that `VERSION` file is used to name the `.so` files, so we need to skip the newline.

merschformann: Ah, makes sense. :+1:

PR #37, opened 2022-08-03
Feature/eng 1176 domain tests

nilsbeck:
# Description
This PR adds `model.MaxInt` and `model.MinInt` to be used as platform independent integers. These are needed for new tests in plugins which will be added in [this PR](https://github.com/nextmv-io/plugins/pull/61)

## Changes
- adds `model/const.go`

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -1,5 +1,15 @@
 package model
 
+import "math/bits"
+
+// Constants for integer bounds.
+const (
+	// MaxInt is the maximum value for an integer.
+	MaxInt int = (1<<bits.UintSize)/2 - 1
+	// MinInt is the minimum value for an integer.
+	MinInt = (1 << bits.UintSize) / -2
+)
+
 // A Domain of integers.
 type Domain interface {
 	// Add values to a domain.
--- a/.github/workflows/go-build-test.yml
+++ b/.github/workflows/go-build-test.yml
@@ -38,7 +38,8 @@ jobs:
         run: git config --global url."git@github.com:".insteadOf "https://github.com/"
 
       - name: export sdk version
-        run: echo "VERSION=$(cat VERSION)" >> $GITHUB_ENV
+        run: echo "VERSION=$(go run cmd/version/main.go)" >> $GITHUB_ENV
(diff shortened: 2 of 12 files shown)


larsbeck (dismissed):

model/const.go
```diff
@@ -0,0 +1,18 @@
...
+// related proprietary information is strictly prohibited without the express
+// written permission of nextmv.io, inc.
+
+package model
+
+import "math/bits"
+
+// Constants for integer bounds.
+const (
+	MaxInt int = (1<<bits.UintSize)/2 - 1
```
sebastian-quintero:
```suggestion
	// MaxInt is the maximum value for an integer.
	MaxInt int = (1<<bits.UintSize)/2 - 1
```
Add godoc here plz.

model/const.go
```diff
@@ -0,0 +1,18 @@
...
+	MinInt     = (1 << bits.UintSize) / -2
```
sebastian-quintero:
```suggestion
	// MinInt is the minimum value for an integer.
	MinInt     = (1 << bits.UintSize) / -2
```
Same here bitte.

sebastian-quintero (changes requested):
I don't think that a new file is worthwhile. You can throw these at the top of the existing `model.go` file, up to you.

larsbeck (dismissed):

model/model.go
```diff
@@ -1,5 +1,15 @@
 package model
 
+import "math/bits"
+
+// Constants for integer bounds.
+const (
+	// MaxInt is the maximum value for an integer.
+	MaxInt int = (1<<bits.UintSize)/2 - 1
+	// MinInt is the minimum value for an integer.
+	MinInt = (1 << bits.UintSize) / -2
```
merschformann, line 10: Funny how the formatter thinks one of the two is long enough to compress `<<` and for the other it's not. :joy_cat:

merschformann (dismissed):
:rocket:

sebastian-quintero (dismissed):

sebastian-quintero (approved):

merschformann (approved):

PR #39, opened 2022-08-09
Feature/eng 1280 Implement route output option

nilsbeck:
# Description
This PR adds the API for the `Output` option in router to achieve a customized fleet output.

## Changes
- `route/option.go` has now some more data structures


route/option.go
```diff
@@ -328,6 +328,28 @@ func Sorter(
 	return sorterFunc(sorter)
 }
 
+// PlanMarshaller defines an interface that is used to override router's default
+// output when using multiple vehicles.
+type PlanMarshaller interface {
```
dirkschumacher: `Marshaller` -> `Marshaler`

route/option.go
```diff
@@ -328,6 +328,28 @@ func Sorter(
...
+	Marshal(PartialPlan) ([]byte, error)
+}
+
+// VehicleMarshaller defines an interface that is used to override router's
+// default output when using exactly one vehicle.
+type VehicleMarshaller interface {
```
dirkschumacher: Can we call that `PartialVehicleMarshaler` to be inline with the interface

route/option.go
```diff
@@ -328,6 +328,28 @@ func Sorter(
...
+	Marshal(PartialVehicle) ([]byte, error)
```
dirkschumacher: I would suggest we call the function `MarshalJSON` similar to the `Marshaler` interface in `encoding/json`.

dirkschumacher (commented):
Just some quick comments even though still WIP

route/option.go
nilsbeck: was removed in favor of other implementation

route/option.go
nilsbeck: was removed in favor of other implementation

route/option.go
nilsbeck: was removed in favor of other implementation

sebastian-quintero (dismissed):

dirkschumacher (dismissed):

sebastian-quintero (approved):

dirkschumacher (approved):

PR #41, opened 2022-08-11
Support the SDK module being replaced

sebastian-quintero:
# Description

When building the `.so` binaries and replacing `github.com/nextmv-io/sdk` for the local module the version was still being taken as a specific commit. However, to run the tests for `sdk`, the `dep.Version` is being taken as the fallback version. This is evident in [this action](https://github.com/nextmv-io/sdk/runs/7782265834?check_suite_focus=true).

### DIFF
--- a/version.go
+++ b/version.go
@@ -18,11 +18,24 @@ var versionFallback string
 var VERSION = getVersion()
 
 func getVersion() string {
-	bi, _ := debug.ReadBuildInfo()
-	for _, dep := range bi.Deps {
-		if strings.HasPrefix(dep.Path, "github.com/nextmv-io/sdk") {
-			return dep.Version
-		}
+	bi, ok := debug.ReadBuildInfo()
+	if !ok {
+		return versionFallback
 	}
+
+	for _, dep := range bi.Deps {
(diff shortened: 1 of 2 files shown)


version.go
```diff
@@ -18,11 +18,24 @@ var versionFallback string
...
-	bi, _ := debug.ReadBuildInfo()
+	bi, ok := debug.ReadBuildInfo()
+	if !ok {
+		return versionFallback
+	}
+
 	for _, dep := range bi.Deps {
-		if strings.HasPrefix(dep.Path, "github.com/nextmv-io/sdk") {
-			return dep.Version
+		// Get only care about this repo being used as a dependency.
```
merschformann:
```suggestion
		// We only care about this repo being used as a dependency.
```
typo?

mamclaughlin (dismissed):

merschformann (dismissed):
:rocket:

version.go
sebastian-quintero: Thank you.

merschformann (approved):

mamclaughlin (approved):

PR #51, opened 2022-08-24
Feature/ENG-1437: accessing estimated arrival times in the update option

nilsbeck:
# Description

This PR extends the PartialVehicle interface to add a new Times() function that gives access to ETA, ETS and ETD of the vehicle.

# Changes

- Adds the new data to route/`router.go`

### DIFF
--- a/route/router.go
+++ b/route/router.go
@@ -77,18 +77,19 @@ type PartialVehicle interface {
 	// Value return the value of vehicle. Usually this is the cost of the route.
 	Value() int
 	// Times returns a the estimated time of arrival (ETA), estimated time of
-	// service (ETS) and estimated time of departure (ETD) for each stop in the
-	// route. Usually ETA is the same as ETS, unless there is a waiting time
-	// before a time window opens, which is when the service actually starts
-	// (ETS).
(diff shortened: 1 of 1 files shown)


route/router.go
```diff
@@ -76,6 +76,19 @@ type PartialVehicle interface {
...
+	// service (ETS) and estimated time of departure (ETD) for each stop in the
+	// route. Usually ETA is the same as ETS, unless there is a waiting time
+	// before a time window opens, which is when the service actually starts
+	// (ETS).
+	Times() Times
+}
+
+// Times holds the ETA, ETS and ETD information.
+type Times struct {
+	ETA []int `json:"eta"`
```
dirkschumacher: Let's try not to use acronyms and align this with the naming in `Plan`.
See here: https://github.com/nextmv-io/sdk/blob/017b2f5b0a724d062bd649074d3619bef1810989/route/router.go#L99-L100

dirkschumacher (changes requested):

route/router.go
nilsbeck: changed

dirkschumacher (approved):

davidrijsman (approved):

PR #53, opened 2022-08-25
Feature/ENG-1449: Propagate function in store needs better documentation

nilsbeck:
# Documentation

This PR improves the already existing documentation for `store.Propagate`.

## Changes

- store/`store.go` receives improved documentation and example.

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -102,7 +102,7 @@ type Store interface {
 			s := store.New()
 			x := store.NewVar(s, 1)
 			s = s.Propagate(func(s store.Store) []store.Change {
-				if condition {
+				if x.Get(s) <= 1 {
 					return []store.Change{
 						x.Set(2),
 						x.Set(42),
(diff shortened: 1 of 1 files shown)


davidrijsman (dismissed):

store/store.go
```diff
@@ -96,15 +96,19 @@ type Store interface {
...
+		Propagate changes into a Store and is re-invoked until no further
+		changes need to be made and an empty slice of changes is returned.
 
 			s := store.New()
 			x := store.NewVar(s, 1)
 			s = s.Propagate(func(s store.Store) []store.Change {
-				return []store.Change{
-					x.Set(2),
-					x.Set(42),
+				if condition {
```
dirkschumacher: Let's try to make the code actually work.
You could check if `x.Get(s) <= 1` in that `if` statement.

dirkschumacher (changes requested):

store/store.go
nilsbeck: good point. changed this

dirkschumacher (approved):
LGTM

davidrijsman (approved):

PR #55, opened 2022-08-30
Feature/ENG-1448 domains remove allows for using it incorrectly

nilsbeck:
# Description

This PR changes the signature of the `Remove` function for `model.Domain` and `model.Domains` from using variadic to a fixed slice to avoid unexpected behavior. This behavior could occur when _no_ values to be removed where passed at all (per signature of the function this is valid), but hard to detect by the user.

## Changes
- Changes the signatures in model/`model.go`, store/`domain.go` and store/`domains.go`
- Updates `tests` to adhere to new signatures

### DIFF
--- a/model/model.go
+++ b/model/model.go
@@ -35,7 +35,7 @@ type Domain interface {
 	// Min of a domain and a boolean indicating it is nonempty.
 	Min() (int, bool)
 	// Remove values from a domain.
-	Remove(...int) Domain
+	Remove([]int) Domain
 	// Slice representation of a domain.
 	Slice() []int
 	// Value returns an int and true if a domain is singleton.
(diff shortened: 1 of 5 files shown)


model/model.go
```diff
@@ -35,7 +35,7 @@ type Domain interface {
 	// Min of a domain and a boolean indicating it is nonempty.
 	Min() (int, bool)
 	// Remove values from a domain.
-	Remove(...int) Domain
+	Remove([]int) Domain
```
sebastian-quintero, line 38: I left a similar comment in the plugins counterpart, but can't we change it to keep the syntax from line 64?

```go
Remove(int, ...int) Domain
```

And the syntax for `Domains` is the same

```go
Remove(int, ...int) Domains
```

sebastian-quintero (approved):

davidrijsman (approved):

PR #56, opened 2022-08-30
Feature/eng 1243 complete testable sdk godoc examples - old

davidrijsman:


model/example_domain_test.go
```diff
@@ -0,0 +1,618 @@
+package model_test
```
sebastian-quintero, line 1: There are no examples for `Domain`, as you can see from the first picture.

image

The idea is that these mirror the examples already present for `store.Domain`, so they feel the same. As the [Conventions](https://github.com/nextmv-io/plugins#conventions) state:

> To document examples for interface methods, name the test `func Example<Interface>_<method>()`. This will render the examples alongside the interface. Here is an [example from `store`](https://github.com/nextmv-io/sdk/blob/a0dbeac8f5d41e65e8fdb59cbcd679d533b77b41/store/example_domain_test.go#L10-L41):

```go
func ExampleDomain_add() {
	s1 := store.New()
	d := store.Multiple(s1, 1, 3, 5)
	s2 := s1.Apply(d.Add(2, 4))
	fmt.Println(d.Domain(s1))
	fmt.Println(d.Domain(s2))
	// Output:
	// {[{1 1} {3 3} {5 5}]}
	// {[{1 5}]}
}


func ExampleDomain_atLeast() {
	s1 := store.New()
	d := store.NewDomain(s1, model.NewRange(1, 10), model.NewRange(101, 110))
	s2 := s1.Apply(d.AtLeast(50))
	fmt.Println(d.Domain(s1))
	fmt.Println(d.Domain(s2))
	// Output:
	// {[{1 10} {101 110}]}
	// {[{101 110}]}
}


func ExampleDomain_atMost() {
	s1 := store.New()
	d := store.NewDomain(s1, model.NewRange(1, 10), model.NewRange(101, 110))
	s2 := s1.Apply(d.AtMost(50))
	fmt.Println(d.Domain(s1))
	fmt.Println(d.Domain(s2))
	// Output:
	// {[{1 10} {101 110}]}
	// {[{1 10}]}
}
```

You can see that the examples are rendered alongside the `Domain` interface:

image

model/example_domain_test.go
```diff
@@ -0,0 +1,618 @@
...
+	"fmt"
+
+	"github.com/nextmv-io/sdk/model"
+)
+
+func ExampleNewDomain_empty() {
+	domain := model.NewDomain()
+
+	fmt.Println(domain.Empty())
+
```
sebastian-quintero: I will only make this comment once, but it applies to all tests. We should remove the newline before the `// Output:` declaration. This is because the godocs render a blank new line before the closing curly brace that we do not want. I'm copying the text from the rendered godoc so that you can visualize what I mean.
```suggestion
```

```go
package main

import (
	"fmt"

	"github.com/nextmv-io/sdk/model"
)

func main() {
	domain := model.NewDomain()

	fmt.Println(domain.Empty())

}
```

model/example_domain_test.go
```diff
@@ -0,0 +1,618 @@
...
+import (
+	"fmt"
+
+	"github.com/nextmv-io/sdk/model"
+)
+
+func ExampleNewDomain_empty() {
+	domain := model.NewDomain()
+
```
sebastian-quintero: I will only make this comment once, but it applies to all tests. All the tests that are already in place do not introduce newline before the `fmt.Println` statements. Is it ok of we keep consistency here so all the codebase of examples feels similar? I'm appending an [example from the `store` package](https://github.com/nextmv-io/sdk/blob/a0dbeac8f5d41e65e8fdb59cbcd679d533b77b41/store/example_domain_test.go#L10-L41) so that you can see what I mean:

```go
func ExampleDomain_add() {
	s1 := store.New()
	d := store.Multiple(s1, 1, 3, 5)
	s2 := s1.Apply(d.Add(2, 4))
	fmt.Println(d.Domain(s1))
	fmt.Println(d.Domain(s2))
	// Output:
	// {[{1 1} {3 3} {5 5}]}
	// {[{1 5}]}
}


func ExampleDomain_atLeast() {
	s1 := store.New()
	d := store.NewDomain(s1, model.NewRange(1, 10), model.NewRange(101, 110))
	s2 := s1.Apply(d.AtLeast(50))
	fmt.Println(d.Domain(s1))
	fmt.Println(d.Domain(s2))
	// Output:
	// {[{1 10} {101 110}]}
	// {[{101 110}]}
}


func ExampleDomain_atMost() {
	s1 := store.New()
	d := store.NewDomain(s1, model.NewRange(1, 10), model.NewRange(101, 110))
	s2 := s1.Apply(d.AtMost(50))
	fmt.Println(d.Domain(s1))
	fmt.Println(d.Domain(s2))
	// Output:
	// {[{1 10} {101 110}]}
	// {[{1 10}]}
}
```

model/example_domain_test.go
```diff
@@ -0,0 +1,618 @@
...
+	fmt.Println(domain.Len())
+
+	fmt.Println(domain.Contains(3))
+
+	// Output:
+	// 1 true
+	// 5 true
+	// 3
+	// true
+}
```
sebastian-quintero: I personally don't like the idea of naming the tests with `first` and `second`, because the rendered godocs use alphabetical ordering so it ends up looking like this (I know this is for a different type but the comment applies here):

image

Maybe we can actually unify both tests into a single one, to make it more compact and add a description to show the user what is happening.

```suggestion
// Create an empty domain and one with multiple integer values.
func ExampleMultiple() {
	d1 := model.Multiple()
	fmt.Println(domain.Empty())
	
	d2 := model.Multiple(3, 5, 1)
	fmt.Println(d2.Min())
	fmt.Println(d2.Max())
	fmt.Println(d2.Len())
	fmt.Println(d3.Contains(3))
	// Output:
	// true
	// 1 true
	// 5 true
	// 3
	// true
}
```

model/example_domain_test.go
```diff
@@ -0,0 +1,618 @@
...
+	domain := model.NewDomain(model.NewRange(1, 0))
+
+	fmt.Println(domain.Empty())
+
+	// Output:
+	// true
+}
+*/
+
+func ExampleNewDomain_singleRange_second() {
```
sebastian-quintero: In my mind, I don't think we need that many examples to understand how to build a new domain using `NewDomain`. I would use the strategy I presented above to maybe have `func ExampleNewDomain_single` and `func ExampleNewDomain_multiple` ➡️ two examples that showcase how to build domains using this function. Inside each function you can place multiple cases and add some comments so that the user follow through. I thikn this makes it more compact for the current example naming is a bit cryptic, in my opinion:

image

I also think we are ok with a couple of examples...

sebastian-quintero: By the way, the comment here regarding using `first`,  `second` etc. applies here: they don't get rendered in order so it is confusing.

model/example_domains_test.go
```diff
@@ -0,0 +1,492 @@
+package model_test
+
+import (
+	"fmt"
+
+	"github.com/nextmv-io/sdk/model"
+)
+
+func ExampleNewDomains_first() {
```
sebastian-quintero: The same comment here regarding documenting the methods of the interface and what they do by having `ExampleDomains_add`, `ExampleDomains_assign`, etc... This would mirror what is currently on  `store` to have some consistency.

model/example_domain_test.go
```diff
@@ -0,0 +1,618 @@
...
+	fmt.Println(domain.Min())
+
+	fmt.Println(domain.Max())
+
+	// Output:
+	// 0 true
+	// 2 true
+}
+
+func ExampleDomain_Remove_first() {
```
sebastian-quintero: I see, so you did attempt to add examples for the interface methods. However, the naming is wrong. If you render the package docs locally, you can realize that these examples are not displayed anywhere. Please see the [linting and testing](https://github.com/nextmv-io/plugins#linting-and-testing) guide:

> When adding new content to the Go Docs, including testable examples, check the rendered package docs locally. Stand on the root of sdk and run:

```bash
$ pkgsite
2022/08/10 16:32:27 Info: Listening on addr http://localhost:8080
```
> Then open up your browser and go to the desired package. You can start from the module itself:

```text
http://localhost:8080/github.com/nextmv-io/sdk
```

To make the example render follow the suggestion.

```suggestion
func ExampleDomain_remove() {
```

I would apply the same pattern as I have commented out before. Instead of having a `first`, `second` etc... variations, have a single test where you can display multiple domains adding some comments so that the user follows... This is so the number of tests also don't become unmanageable. I will only make this comment once as it applies to the rest of the examples that follow in this file.

model/example_domains_test.go
```diff
@@ -0,0 +1,492 @@
...
+
+	fmt.Println(domains.Len())
+	fmt.Println(domains.Domain(0).Value())
+	fmt.Println(domains.Domain(1).Value())
+
+	// Output:
+	// 2
+	// 1 true
+	// 1 true
+}
```
sebastian-quintero: Here is an example of how to condense the three examples into a single test. I think it is nice and compact and guides the reader to understand what domain they are seeing in the output. This suggestion can be applied to the other cases as well, where the domain "id" is printed alongside the result.

```suggestion
// Domains behave differently when repeating empty or integer domains.
func ExampleRepeat() {
	// An empty domain is repeated 0 times.
	d1 := model.Repeat(0, model.NewDomain())
	fmt.Println("d1: ", d1.Len())
	fmt.Println("d1: ", d1.Empty())
	
	// An empty domain is repeated 1 time.
	d2 := model.Repeat(1, model.NewDomain())
	fmt.Println("d2: ", d2.Len())
	fmt.Println("d2: ", d2.Empty())
	
	// A non-empty domain is repeated multiple times.
	d3 := model.Repeat(2, model.Singleton(1))
	fmt.Println("d3: ", d3.Len())
	fmt.Println("d3: ", d3.Domain(0).Value())
	fmt.Println("d3: ", d3.Domain(1).Value())
	// Output:
	// d1: 0
	// d1: true
	// d2: 1
	// d2: true
	// d3: 2
	// d3: 1 true
	// d3: 1 true
}
```

model/example_domain_test.go
```diff
@@ -0,0 +1,618 @@
...
+
+	fmt.Println(domain.Empty())
+
+	fmt.Println(domain.Value())
+
+	fmt.Println(domain.Min())
+
+	fmt.Println(domain.Max())
+
+	iterator := domain.Iterator()
```
sebastian-quintero: Commenting here is weird but applies because you are using `Iterator()` here. I believe a test for the `type Iterator` is missing, so it would be good to add that.

route/example_measure_test.go
```diff
@@ -0,0 +1,94 @@
...
+	// Output:
+	// 1.234
+	// 1.234
+	// 4.321
+}
+
+func ExampleHaversineByPoint() {
+	byPoint := route.HaversineByPoint()
+
+	measure := byPoint.Cost(route.Point{1, 2}, route.Point{4, 5})
```
sebastian-quintero: Can we make these coordinates something more real? That way the output is not extremely large and people understand these are points on the Earth, not indices.

route/example_measure_test.go
```diff
@@ -0,0 +1,94 @@
...
+	fmt.Println(int(measure))
+
+	// Output:
+	// 471293
+}
+
+func ExampleConstantByPoint() {
+	byPoint := route.ConstantByPoint(1.234)
+
+	measure := byPoint.Cost(route.Point{1, 2}, route.Point{4, 5})
```
sebastian-quintero: Same here, can I haz real coordinates?

route/example_measure_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	// Output:
+	// 1.234
+}
+
+func ExampleIndexed() {
+	haversineByPoint := route.HaversineByPoint()
+
+	points := []route.Point{
+		{1, 2},
```
sebastian-quintero: Same comment regarding real coordinates.

sebastian-quintero (changes requested):
Bunch of comments 😅

davidrijsman:
@sebastian-quintero had to rethink this, it's not testing but examples with tests on the examples. Naming convention does not always result in examples appearing in doc so changed to draft until we can figure out what is going on.

model/example_domain_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+func ExampleDomain_empty() {
+	d1 := model.NewDomain()
+	d2 := model.Singleton(42)
+	fmt.Println(d1.Empty())
+	fmt.Println(d2.Empty())
+	// Output:
+	// true
+	// false
+}
```
sebastian-quintero, line 73: We are missing `func ExampleDomain_iterator` to demonstrate use of `Domain.Iterator`.

model/example_domains_test.go
```diff
@@ -0,0 +1,196 @@
...
+func ExampleDomains_domain() {
+	d := model.NewDomains(model.NewDomain(), model.Singleton(42))
+	fmt.Println(d.Domain(0))
+	fmt.Println(d.Domain(1))
+	// Output:
+	// {[]}
+	// {[{42 42}]}
+}
+
+func ExampleDomains_domains() {
```
sebastian-quintero: `Domains` is not a method on the `Domains` interface, only `Domain` is a method. I think you can delete this method or rename it to `ExampleDomains` so that it is a general example on the `Domains` interface.

route/example_measure_test.go
sebastian-quintero: @davidrijsman This is still not acknowledged.

route/example_measure_test.go
sebastian-quintero: @davidrijsman This is still not acknowledged.

route/example_measure_test.go
sebastian-quintero: @davidrijsman This is still not acknowledged.

sebastian-quintero (changes requested):
Some small things that are still leftover, in my opinion.

model/example_domain_test.go
davidrijsman: Added

model/example_domains_test.go
davidrijsman: Removed

sebastian-quintero (dismissed):

dirkschumacher (dismissed):

sebastian-quintero (approved):

dirkschumacher (approved):

PR #58, opened 2022-09-08
Add support for build tag "debug"

larsbeck:
## Description

We are building plugins with and without debug flags. This PR adds support to use one or the other version as a user of SDK. When a user wants to debug, they provide `-tags=debug` as their `buildflags` for example in their VsCode launch configuration (this will already be set in the workspace files of `nextmv-cli` templates).

### DIFF
--- a/plugin/connect.go
+++ b/plugin/connect.go
@@ -83,12 +83,13 @@ func potentialPluginPaths(slug string) ([]string, error) {
 
 	// Assemble potential plugin paths
 	filename := fmt.Sprintf(
-		"nextmv-%s-%s-%s-%s-%s.so",
+		"nextmv-%s-%s-%s-%s-%s%s.so",
 		slug,
 		sdk.VERSION,
 		runtime.Version(),
 		runtime.GOOS,
 		runtime.GOARCH,
+		debug,
 	)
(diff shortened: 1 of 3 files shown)


nilsbeck (approved):

plugin/connect.go
```diff
@@ -83,12 +83,13 @@ func potentialPluginPaths(slug string) ([]string, error) {
 
 	// Assemble potential plugin paths
 	filename := fmt.Sprintf(
-		"nextmv-%s-%s-%s-%s-%s.so",
+		"nextmv-%s-%s-%s-%s-%s%s.so",
```
sebastian-quintero, line 86: Please see my comment on the private implementation. Not a blocker but maybe we can keep consistency of arguments being separated by a dash. Not a blocker though.
```suggestion
		"nextmv-%s-%s-%s-%s-%s-%s.so",
```

sebastian-quintero (approved):
Small non-blocking comment. 🌮

plugin/connect.go
larsbeck: It is separated by a dash, but the dash comes in on the variable.

PR #77, opened 2022-09-22
Simplify plugin loading

larsbeck:
# Description

This PR simplifies how method definitions are mapped to their implementations in plugins. It also makes the mapping more efficient because it only maps the functions needed rather than all functions.
With this PR it becomes simpler to add more methods because there is only one place where you need to add a mapping call and that is where the method to be mapped is defined.

## Changes
- Added a `Connector` which is the central place holding convention base mapping logic
- Packages now use the `Connector` and usage of passing `string`s has been minimized

Resolves ENG-1600


connect/connect.go
```diff
@@ -0,0 +1,61 @@
+// Package connect provides a Connector which allows to connect method
+// definition with their implementations in plugins
+package connect
```
nilsbeck, line 3: Can the package be called connector? then it's not `connect.connect(...)` but `connector.connect(...)`

connect/connect.go
```diff
@@ -0,0 +1,61 @@
...
+	if _, ok := c.connected[target]; ok {
+		return
+	}
+
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+
+	if _, ok := c.connected[target]; ok {
+		return
+	}
```
nilsbeck, line 41: Do we need this twice?

connect/connect.go
```diff
@@ -0,0 +1,61 @@
...
+
+	// get the calling function, get ok to make linter happy
```
nilsbeck, line 43: 😃

store/variable.go
```diff
@@ -36,6 +38,7 @@ NewVar stores a new variable in a Store.
 	x := store.NewVar(s, 10) // x is stored in s.
 */
 func NewVar[T any](s Store, data T) Var[T] {
+	connect.Connect(con, &newVarFunc)
```
nilsbeck, line 41: why do we need to connect here now and before not?

nilsbeck (commented):
Trying to understand, I have a few questions....

connect/connect.go
```diff
@@ -0,0 +1,61 @@
...
+	"fmt"
+	"runtime"
+	"strings"
+	"sync"
+
+	"github.com/nextmv-io/sdk/plugin"
+)
+
+// Connector connects methods with their implementations in plugins.
+type Connector struct {
```
sebastian-quintero, line 15:
```suggestion
type connector struct {
```
Should this be unexported? I don't see it being actively referenced as `connect.Connector` anywhere outside this package, right?

sebastian-quintero (dismissed):
Non-blocker 🌮

connect/connect.go
larsbeck: It is used in all packages that use the Connector, so we can't unexport. We could introduce an interface instead, but I don't see value in that at this time.

store/variable.go
larsbeck: Before every call to `connect()` would connect all components. Well only the first call would connect them all and all subsequent calls would be noops.

connect/connect.go
larsbeck: The linter doesn't like when you ignore more than two return values in its current setting...

connect/connect.go
larsbeck: Yes, the first call is a performance optimization. Then you lock. When the lock is released the second waiting goroutine, when it passes the lock, doesn't need to do all the code after that check again. You could do all this without a lock. It was like that and is fine from my point of view, so I left it that way.

connect/connect.go
larsbeck: There is a `Connector` in `connect`, like there is a `Router` in `route`, so I think this is in line with how we name things. Only difference is, that there is a `Connect` method (which needs to be there for technical reasons required by how Go handles Generics - happy to explain live) whereas `route` does not need a `Route` method.

nilsbeck (approved):

connect/connect.go
sebastian-quintero: Yeah but you're using `Connect` and `NewConnector`, right? You're not referencing the `Connector` type directly, that's why I suggested having the type be private, whilst of course keeping the functions public.

sebastian-quintero (approved):

connect/connect.go
larsbeck: The linter doesn't like that because the returned type would be unexported.

PR #85, opened 2022-09-27
Refactor MIP naming conventions and API

davidrijsman:
NewBinaryVar, NewContinuousVar, NewIntegerVar, NewConstraint should no longer return an error
Allow the user to override the default name of a variable and constraint (Name(), SetName(string))
Allow to query for a term in a constraint. Constraint::Term(Var) Term
Allow to query for a term in  the objective. Objective::Term(Var) Term
Printing terms of a constraint should always be done in order of index of the associated variable
Printing terms of an objective should always be done in order of index of the associated variable
Change API SetVerboseLevel to SetVerbosity
Rename IntegerVar to Int, this implies NewIntegerVar becomes NewInt
Rename BinaryVar to Bool, this implies NewBinaryVar becomes NewBool
Rename ContinuousVar to Float,  this implies NewContinuousVar becomes NewFloat


larsbeck:
If I understood @ryanjoneil correctly [here](https://linear.app/nextmv/issue/ENG-1596#comment-ff6c5291) he thinks that querying `Terms()` (returning a slice, see: "...in string representations and **slices** returned to the user...") should also return terms in the same order. This came up previously where we decided to fix the test instead of `Terms()` but I now second that opinion. Maybe there is a way to have this be in the same order without the runtime penalty (which was the reason not to immediately fix it).

davidrijsman:
> 

Currently you can get the terms in any order you like by sorting the output of Terms. Terms query itself needs to stay O(n) on number of terms. To make the Stringer deterministic they will be sorted on index for the variables, that makes the Stringer O(n log n) but that is only invoked when stringer is used obviously.

So if you need the terms in order of index of the variable a user can do:

```
terms := entity.Terms()
sort.SliceStable(terms, func(i, j int) bool {
      return terms[i].Var().Index() < terms[j].Var().Index()
})
```
If you need the terms in order of the name of the variable a user can do:
```
terms := entity.Terms()
sort.SliceStable(terms, func(i, j int) bool {
      return strings.Compare(terms[i].Var().Name(), terms[j].Var().Name())
})
```

Obviously if we can get Terms in a deterministic order without any additional cost we are also good.

ryanjoneil:
> To make the Stringer deterministic they will be sorted on index for the variables, that makes the Stringer O(n log n) but that is only invoked when stringer is used obviously.

I think that's OK. A user shouldn't normally be string-ifying models willy nilly. I'd be OK with that performance penalty on `String()` if the modeling APIs aren't penalized.

mip/example_constraint_test.go
```diff
@@ -38,22 +38,24 @@ func ExampleConstraint_equal() {
 func ExampleConstraint_lessThanOrEqual() {
 	model := mip.NewModel()
 
-	c, _ := model.NewConstraint(mip.LessThanOrEqual, 1.0)
-
+	c := model.NewConstraint(mip.LessThanOrEqual, 1.0)
+	c.SetName("constraint")
```
ryanjoneil, line 42: Just to be clear, calling `SetName` isn't _required_, right? I'm not sure if these examples show up in the docs or not. If they do, I think we should only call `SetName` in a single example showing how to use the string representation of a model.

ryanjoneil (commented):
Coming along quite nicely.

mip/example_constraint_test.go
davidrijsman: It is not required, it reverts to what it already was (and now I realize a Float should revert to F{index} and not C{index})

sebastian-quintero (approved):

dirkschumacher (approved):

PR #89, opened 2022-09-30
ENG-1678 Hide private output

sebastian-quintero:
# Description

Hide the output of a bash script.

### DIFF
--- a/.github/workflows/go-build-test.yml
+++ b/.github/workflows/go-build-test.yml
@@ -30,8 +30,7 @@ jobs:
 
       # Gets the token and stores it in a variable.
       - name: export local branch
-        run:
-          echo "BRANCH_CHECK=$(git ls-remote --heads https://github.com/nextmv-io/plugins ${{ env.LOCAL_BRANCH }})" >> $GITHUB_ENV
+        run: echo "BRANCH_CHECK=$(git ls-remote --heads https://github.com/nextmv-io/plugins ${{ env.LOCAL_BRANCH }})" >> $GITHUB_ENV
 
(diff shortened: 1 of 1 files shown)


dirkschumacher (approved):

.github/workflows/go-build-test.yml
```diff
@@ -72,7 +71,7 @@ jobs:
           go mod edit -replace=github.com/nextmv-io/sdk=${{env.SDK_PATH}}
 
           # Build the plugins.
-          REMOVE_TRIMPATH=1 NEXTMV_LIBRARY_PATH=${{ env.NEXTMV_LIBRARY_PATH }} bash scripts/build.sh
+          REMOVE_TRIMPATH=1 NEXTMV_LIBRARY_PATH=${{ env.NEXTMV_LIBRARY_PATH }} bash scripts/build.sh > /dev/null 2>&1
```
merschformann, line 74:
```suggestion
          REMOVE_TRIMPATH=1 NEXTMV_LIBRARY_PATH=${{ env.NEXTMV_LIBRARY_PATH }} bash scripts/build.sh &> /dev/null
```

This is equivalent to this, correct? I am used to doing this, but I think I get how yours works. :thinking: 
Anyway, just curious not at all blocking.
Also, I get it right that this means to silence stdout & stderr, right?

merschformann (approved):
:rocket:

.github/workflows/go-build-test.yml
sebastian-quintero: Yeah the idea is to silence all output.

PR #113, opened 2022-10-20
Add missing measure features

sebastian-quintero:
# Description

Add missing measure features (`route.ScaleByPoint`, `route.IsTriangular`) and rework documentation to include testable examples as opposed to code snippets in the godocs.

### DIFF
--- a/route/measure.go
+++ b/route/measure.go
@@ -121,7 +121,7 @@ func Scale(m ByIndex, constant float64) ByIndex {
 
 // ScaleByPoint scales the cost of some other measure by a constant.
 func ScaleByPoint(m ByPoint, constant float64) ByPoint {
-	connect.Connect(con, &scaleFunc)
+	connect.Connect(con, &scaleByPointFunc)
 	return scaleByPointFunc(m, constant)
 }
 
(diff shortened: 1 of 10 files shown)


route/measure.go
```diff
@@ -184,6 +184,14 @@ func Matrix(arcs [][]float64) ByIndex {
 	return matrixFunc(arcs)
 }
 
+// IsTriangular returns true if the triangle inequality holds for the provided
+// measure. It returns false if the measure does not implement the Triangular
+// interface or the triangle inequality does not hold.
+func IsTriangular(m any) bool {
```
dirkschumacher, line 190: Do we really need this function when we have the Triangular interface? Why don't we require the type to be at least `func IsTriangular(m Triangular) bool`

dirkschumacher (approved):
One comment/question

route/measure.go
sebastian-quintero: The reason is that the method checks if the measure `m` implements the interface, so if you know before the fact that a measure implements that interface you would not need this method. That is how I understand it at least.

nmisek (approved):

PR #116, opened 2022-10-31
Bugfix/eng 1795 investigate why runs are passing cloud fleet but failing router

nilsbeck:
# Description

This PR fixes the missing initialization costs in the `cloud-routing` template

## Changes
- `cloud-routing` template now uses init costs in the overridden value function


larsbeck (approved):

merschformann (approved):
:rocket:

PR #117, opened 2022-11-01
Feature/eng 1793 move data frame code into sdk plugin

davidrijsman:
Introduction of data frames.


dataframe/aggregation.go
```diff
@@ -0,0 +1,33 @@
+package dataframe
+
+import "fmt"
+
+// Aggregation defines how to aggregate rows of a group of rows in a Groups
+// instance.
+type Aggregation interface {
+	fmt.Stringer
+
+	// Column return the column the aggregation will be applied to.
```
larsbeck:
```suggestion
	// Column returns the column the aggregation will be applied to.
```

dataframe/aggregation.go
```diff
@@ -0,0 +1,33 @@
...
+	Column() Column
+
+	// As return the column  to be used to identify the newly created column
```
larsbeck:
```suggestion
	// As returns the column to be used to identify the newly created column.
```

dataframe/dataframe.go
```diff
@@ -0,0 +1,45 @@
...
+// data manipulation.
+type DataFrame interface {
+	// Column returns a column identified by name, panics if not present.
+	Column(name string) Column
+	// Columns returns all columns present in the dataframe.
+	Columns() Columns
+
+	// Distinct returns a new DataFrame that only contains unique rows with
+	// respect to the specified columns. If no columns are given Distinct will
+	// return rows where allow columns are unique.
```
larsbeck:
```suggestion
	// return rows where all columns are unique.
```
Maybe?

dataframe/dataframe.go
```diff
@@ -0,0 +1,45 @@
...
+	Distinct(columns ...Column) DataFrame
+
+	// Filter returns a new filtered DataFrame according to the filter.
+	Filter(filter Filter) DataFrame
+
+	// GroupBy groups rows together for which the values of specified columns
+	// are the same.
+	GroupBy(columns ...Column) Groups
```
larsbeck, line 21: It would be good to see an example as `GroupBy` on multiple columns can be understood in many ways - does ordering matter for example? Or is it random?

larsbeck (dismissed):
Nice work! Do we plan to have usage examples or docs?

dataframe/dataframe.go
davidrijsman: GroupBy ordering does not matter (no rows dissapear so not sure how ordering could matter?). I agree examples would be a good thing, documentation will be in next release.

store/example_store_test.go
```diff
@@ -147,13 +147,10 @@ func ExampleDefaultOptions() {
 	// {
 	//   "diagram": {
 	//     "expansion": {
-	//       "limit": 0
+	//       "limit": 1
```
dirkschumacher: why did that change?

dataframe/column.go
```diff
@@ -0,0 +1,130 @@
...
+// BoolColumn is the typed column of type Bool.
+type BoolColumn interface {
+	Column
+
+	// IsFalse creates a filter to filter all rows having value false.
+	IsFalse() Filter
+	// IsTrue creates a filter to filter all rows having value true.
+	IsTrue() Filter
+
+	// Value return the value at row for dataframe df,
```
dirkschumacher, line 73:
```suggestion
	// Value returns the value at row for dataframe df,
```

dirkschumacher (approved):

larsbeck (approved):

PR #119, opened 2022-11-03
feature/ENG-1770 embed template and input file content

racquesta:
Adds templates and inputs packages with exported variables holding embedded content from template and sample input files.

### DIFF
--- a/templates/templates.go
+++ b/templates/templates.go
@@ -2,7 +2,8 @@
 package templates
 
 import (
-	// Embed ...
+	// This package is required to embed files using the `//go:embed` directive albeit
+	// it is not used directly.
 	_ "embed"
 )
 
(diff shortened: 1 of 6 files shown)


nmisek (dismissed):

mamclaughlin (dismissed):

templates/templates.go
```diff
@@ -0,0 +1,171 @@
+// Package templates contains variables holding embded template files.
+package templates
+
+import (
+	// Embed ...
```
sebastian-quintero: We can also help a future Nextmver understand this by saying something like this.
```suggestion
	// This package is required to embed files using the `//go:embed` directive albeit
	// it is not used directly.
```

sebastian-quintero (dismissed):

nmisek (dismissed):

sebastian-quintero (approved):

mamclaughlin (approved):

PR #121, opened 2022-11-07
feature/ENG-1390 market spend template

nilsbeck:
# Description

Adds a budget allocation MIP template to cli.

## Changes

* new template folder for mip-budget-allocation.


nmisek:
@nilsbeck - before we release this - (which may be tomorrow if this slips in?) please complete this Notion page for the template so ops is prepared to talk about it when it releases https://www.notion.so/nextmv/Nextmv-models-apps-and-templates-3baa6fd7f16f4ba08bc5d5359b346ac3?p=ad311dad6d6145c48925a5bc88b260fa&pm=s

nmisek:
@nilsbeck from Tiff / Haley - can we rename this template to `mip-incentive-allocation`

templates/mip-incentive-allocation/README.md
```diff
@@ -0,0 +1,43 @@
...
+This template demonstrates how to solve a Mixed Integer Programming problem
+using the open-source solver [HiGHS](https://github.com/ERGO-Code/HiGHS). To run
+this beta template, please contact support@nextmv.io for access to the Nextmv
+HiGHS interface.
+
+To solve a Mixed Integer Problem (MIP) is to optimize a linear objective
+function of many variables, subject to linear constraints. We demonstrate this
+by solving a incentive allocation problem.
+
+Given a collection of incentives per a user with a cost and effect, our
```
larsbeck:
```suggestion
Given a collection of incentives per user with a cost and effect, our
```

larsbeck (dismissed):
🚀

larsbeck (dismissed):

.golangci.yml
```diff
@@ -197,7 +197,7 @@ issues:
     - path: templates/mip-knapsack/main\.go
       linters:
         - gocyclo
-    - path: templates/mip-meal-allocation/main\.go
+    - path: templates/mip-incentive-allocation/main\.go
```
sebastian-quintero: The idea for _new_ templates, at least how I view it, is to lint them "appropriately" and not waive the linter anymore to avoid accruing tech debt.

sebastian-quintero (dismissed):

.golangci.yml
nilsbeck: I don't think that there is a sensible way of reducing cyclomatic complexity in this case (and likely also not in the other templates). An effort of changing it would either separate the model definition into different functions. That would make it hard to read and understand. Or creating functions that pass the store and everything around it in a not very nice way.
I think, because of how the store api is constructed and how we use the scope of variables with it, it is likely that we often have longer code snippets.
A cyclomatic complexity of `10` is certainly a good guidance - but also a random number.

sebastian-quintero: Agree to disagree 😅. I think we can extract the contents of the `root.Format` function to a separate `func format(s store.Store) any {...}` and then do `root = root.Format(format)`. That formatting I don't see as quite core to the `main` logic. How I see it, the `main` func is an orchestrator and having all the logic there is not necessary. If I'm reading the code I understand it as "oh ok, so they format this `root` thing here, fine", and I don't need to read the contents of that `format` func immediately. The cyclo complexity of 10 is set by @ryanjoneil so I don't think it is very effective if we keep waiving it. I would like your guidance here @ryanjoneil @nmisek 🙏🏻.

larsbeck: I am a little torn on this one. We can factor out the formatting, but it is a little more than just `root = root.Format(format)` because of the scoped `so` (solution) variable needs to get passed to the `format` function as well. And I believe that is what Nils is getting at. We use scoping extensively, which leads to long nested code blocks. So if we want to keep complexity low (as measured by the cyclo complexity linter) then we need to move (parts of the code at least) to passing parameters.

nilsbeck: yes. I think in this simple mip case factoring out formatting will likely be enough. But for "real" store models it will not, I believe. And in those cases factoring it out will get a bit more messy and not necessarily easier to read/follow.
And I think all templates should follow the same structure, this will make it easier to read one after another because things look familiar.
I raised my concern. So, whatever we want to do here, I am all in for it.

nilsbeck: P.S.: I think it is nice that you can read them top to bottom without jumping :)

nmisek: @nilsbeck @sebastian-quintero @larsbeck may i propose that we follow Sebas' suggestion to extract the Format contents here? I understand that alone won't fix complexity of all templates going forward - but I would say let's create a spike to investigate whether this is feasible for _existing_ templates and treat that as a separate tech debt task to decide how we want to handle cyclomatic complexity. Does that work?

nilsbeck: sure, but I guess we should make that same change for all templates then and update the docs accordingly

nilsbeck: I created a ticket to make the changes to the other tickets as well: https://linear.app/nextmv/issue/ENG-1998/refactor-templates-and-pull-the-format-function-out

larsbeck (approved):

nmisek (approved):

PR #136, opened 2022-11-30
General runners

larsbeck:
# Component Visualization

## The `GenericRunner`

It shows the components of a runner. It is used internally by the `CliRunner` and the `HTTPRunner`.
The main components are:

- The `IOProducer` is responsible for providing the input and option sources as well as the output sink for the rest of the pipeline getting executed with each run. In the case of the `HTTPRunner` (see below) the input source is a request body (`io.Reader`) and the option source is `url.Values`. The output is either the http response (`io.Writer`) or, in the async case, a `bytes.Buffer` which will be the content of an http POST to some user-defined callback URL.
- The `InputDecoder` knows how to read from the input source and turn that data into a user-defined input structure
- The `OptionDecoder` knows how to read from the option source and turn that data into a user-defined option structure
- The `Algorithm` is some user-defined algorithm
- The `Encoder` knows how to turn a user-defined solution structure into data that can be written to the output as defined by the `IOProducer`

GenericRunner

## The `CliRunner`

It is a runner which comes with the following main features:

- support to parse (runner- and algorithm-related) options from flags and env vars
- take one input from `Stdin` or a file and write the solutions to `Stdout` or a file
- support for (de-)compression handling
- populate flags so that algorithm options will be parsed automatically and populate defaults if they are given via Go's struct tags.
- decoding of inputs and encoding of the output is configurable and extensible. Currently supported formats are `JSON` and `CSV`.
- Output format/structure is defined by the user's algorithm output. By default it gets enriched with some metadata, but this can be turned off via the `runner.output.quiet` option.

CliRunner

## The `HTTPRunner`

The `HTTPRunner` has all of the features of the `CliRunner`. The main differences / additional features are:

- Stands up an http server (and serves requests until it gets shut down)
- Each request will result in one execution of the algorithm if the request is not rejected; this is a major difference to the `CliRunner` - multiple requests potentially run in parallel
- Built in and configurable rate limiting will limit the number of concurrent algorithm runs
- Gets input from http request body (can be changed by the user)
- Gets algorithm option from http query params and overrides the algorithm option parsed via flags/env vars (can be changed by the user)
- Logging sink can be configured
- http middleware can be plugged in
- A user can configure an http POST endpoint to get results posted to it asynchronously

HTTPRunner

## Tests

Tests are in `Docs2`


run/http_runner.go
```diff
@@ -0,0 +1,264 @@
...
+			algorithm,
+			GenericEncoder[Solution, Option](encode.JSON()),
+		),
+	}
+
+	runnerConfig := runner.Runner.RunnerConfig()
+	runner.maxParallel = make(chan struct{}, runnerConfig.Runner.HTTP.MaxParallel)
+
+	// default http server
+	runner.httpServer = &http.Server{
```
dirkschumacher, line 115: You need to set a `ReadHeaderTimeout` here, otherwise the server is vulnerable to the [Slowloris Attack](https://medium.com/a-journey-with-go/go-understand-and-mitigate-slowloris-attack-711c1b1403f6).
Flagged by `gosec`: 
```
G112 (CWE-400): Potential Slowloris Attack because ReadHeaderTimeout is not configured in the http.Server (Confidence: LOW, Severity: MEDIUM)
```

go.mod
```diff
@@ -1,3 +1,11 @@
 module github.com/nextmv-io/sdk
 
 go 1.19
+
+require (
+	github.com/google/uuid v1.3.0
+	github.com/gorilla/schema v1.2.0
```
dirkschumacher, line 7: https://github.com/gorilla/schema is not maintained anymore

run/cli_ioproducer.go
```diff
@@ -0,0 +1,35 @@
...
+	"context"
+	"io"
+	"log"
+	"os"
+)
+
+// CliIOProducer is the IOProducer for the CliRunner. The input and output paths
+// are used to configure the input and output readers and writers. If the paths
+// are empty, os.Stdin and os.Stdout are used.
+func CliIOProducer(_ context.Context, cfg CLIRunnerConfig) IOData {
```
dirkschumacher: Why not `CLIOProducer`? A bit less readable I admit, but consistent to the other naming. Your call 😅

run/cli_ioproducer.go
```diff
@@ -0,0 +1,35 @@
...
+	reader := os.Stdin
+	if cfg.Runner.Input.Path != "" {
+		r, err := os.Open(cfg.Runner.Input.Path)
+		if err != nil {
+			log.Fatal(err)
```
dirkschumacher: I wonder if `log.Fatal` here is the right thing to do. That means the user of that function cannot recover from a panic and no `defer` calls are executed. Applies to other `log.Fatal` calls as well.

run/cli_ioproducer.go
```diff
@@ -0,0 +1,35 @@
...
+	"os"
+)
+
+// CliIOProducer is the IOProducer for the CliRunner. The input and output paths
+// are used to configure the input and output readers and writers. If the paths
+// are empty, os.Stdin and os.Stdout are used.
+func CliIOProducer(_ context.Context, cfg CLIRunnerConfig) IOData {
+	reader := os.Stdin
+	if cfg.Runner.Input.Path != "" {
+		r, err := os.Open(cfg.Runner.Input.Path)
```
dirkschumacher, line 15: Whose responsible is it to close the file?

run/cli_ioproducer.go
```diff
@@ -0,0 +1,35 @@
...
+		if err != nil {
+			log.Fatal(err)
+		}
+		reader = r
+	}
+	var writer io.Writer = os.Stdout
+	if cfg.Runner.Output.Path != "" {
+		w, err := os.Create(cfg.Runner.Output.Path)
```
dirkschumacher, line 23: Whose responsible is it to close the file?

dirkschumacher (commented):
Great work, really looking forward to the new runners. Some early comments already, not done with the review yet 😬

run/runner.go
```diff
@@ -1,21 +1,46 @@
...
+	GetEncoder() Encoder[Solution, Option]
+	// RunnerConfig returns the runnerConfig of a runner.
+	RunnerConfig() RunnerConfig
+}
+
+// IOProducer is a function that produces the input, option and writer.
+type IOProducer[RunnerConfig any] func(context.Context, RunnerConfig) IOData
+
+// Decoder is a function that decodes a reader into a struct.
+type Decoder[Input any] func(context.Context, any) (Input, error)
```
dirkschumacher, line 31: What is the reason the last element is `any` vs `io.Reader` for example?

run/decoder.go
```diff
@@ -0,0 +1,85 @@
...
+	decoder decode.Decoder
+}
+
+// Decoder is a function that decodes the input from the reader. It uses the
+// given decoder to decode the input. If the input is gzipped, it will be
+// decoded using the gzip.reader.
+func (g *genericDecoder[Input]) Decoder(
+	_ context.Context, reader any) (input Input, err error,
+) {
+	ioReader, ok := reader.(io.Reader)
```
dirkschumacher, line 44: We should say in the docs that we expect the reader to implement `io.Reader`

run/encoder.go
```diff
@@ -0,0 +1,107 @@
...
+// Encode encodes the solution using the given encoder. If a given output path
+// ends in .gz, it will be gzipped after encoding.
+func (g *genericEncoder[Solution, Options]) Encode(
+	_ context.Context,
+	solutions <-chan Solution,
+	writer any,
+	runnerCfg any,
+	options Options,
+) error {
+	ioWriter, ok := writer.(io.Writer)
```
dirkschumacher, line 57: We should also document that it needs to be an `io.Writer`

run/cli_ioproducer.go
larsbeck: The linter was complaining about "unknown word", which isn't terrible, but that was one reason not to capitalize. The other reason was, as you say, readability. Any strong opinions? @ryanjoneil @merschformann @davidrijsman @nilsbeck ?

run/cli_ioproducer.go
larsbeck: The `Decoder`, good catch. I'll add the missing code.

run/cli_ioproducer.go
larsbeck: The `Encoder`, good catch. I'll add the missing code.

run/http_runner.go
larsbeck: Thanks! I'll put in a fix.

run/cli_ioproducer.go
nilsbeck: Hail to the linter! (I guess? 🤷 )

dirkschumacher (commented):
More comments. Forgot to mark the review as finished yesterday 😬

go.mod
dirkschumacher: We may want to fork it and maintain it ourselves in case we want to still use the dependency

run/decoder.go
larsbeck: We don't. It is an implicit contract between the `IOProducer` and the `Decoder`. One can write an `IODecoder` which doesn't pass an `io.Reader` but rather a `string` and the appropriate `Decoder` knows how to decode that `string.`

run/runner.go
larsbeck: Because it doesn't have to be an `io.Reader`. There is no limit as to what an `IOProducer` can provide as the "data source". The `Decoder` will have to be able to handle it though (see comment below).

run/encoder.go
larsbeck: Same as above.

run/cli_ioproducer.go
larsbeck: I introduced an `error` on the `IOProducer` so that users can handle that error in their user code.

go.mod
larsbeck: As discussed, we accept that this is an unmaintained repo. If we encounter issues, we fork it and fix them.

run/cli_ioproducer.go
merschformann: I think "unknown word" comes from the spellcheck extension, right? At least, for me that is the case. So, it's not a linter complaint, I think.

larsbeck: Oh yeah, that is possible. It was tooling yelling at me 😆

run/http_runner.go
```diff
@@ -0,0 +1,265 @@
...
+	// control mechanism to let the request by run async or not.
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer func() { <-h.maxParallel }()
+		// configure how to turn the request and response into an IOProducer.
+		callbackFunc, producer, err := h.httpRequestHandler(w, req)
+		async := callbackFunc != nil
+		if err != nil {
+			handleError(h.httpServer.ErrorLog, async, err, w)
```
dirkschumacher, line 207: There is no `wg.Done()` call here. Doesn't that end up in a permanent block if this error happens?

run/http_runner.go
```diff
@@ -0,0 +1,265 @@
...
+			return
+		}
+		// generate a new requestID
+		requestID := uuid.New().String()
+
+		// get content type from the encoder
+		contentTyper, ok := h.Runner.GetEncoder().(ContentTyper)
+		if !ok {
+			handleError(h.httpServer.ErrorLog, async,
```
dirkschumacher, line 217: There is no `wg.Done()` call here. Doesn't that end up in a permanent block if this error happens?

run/http_request_handlers.go
```diff
@@ -0,0 +1,120 @@
...
+		// Set the GUID header
+		callbackReq.Header.Set("request_id", requestID)
+		// Set the encoding header
+		callbackReq.Header.Set("Content-Type", contentType)
+		// Send the request
+		resp, err := a.httpClient.Do(callbackReq)
+		if err != nil {
+			return err
+		}
+		defer func() {
```
dirkschumacher: We are at the end of the function. No need for a defer I believe

run/generic_runner.go
```diff
@@ -0,0 +1,219 @@
...
+	if retErr != nil {
+		return retErr
+	}
+
+	// handle memory profile
+	deferFuncMemory, retErr := r.handleMemoryProfile(r.runnerConfig)
+	if retErr != nil {
+		return retErr
+	}
+	defer func() {
```
dirkschumacher, line 168: We are already at the end of the function. There is no need for a `defer` call I think

dirkschumacher (commented):
Some more minor comments.

run/generic_runner.go
larsbeck: The alternative code without `defer` should be

```go
	retErr = <-errs
	err = deferFuncMemory()
	if retErr == nil {
		retErr = err
	}

	return retErr
```

but I am uncertain whether that makes the intention more clear.

nmisek (dismissed):

run/generic_runner.go
dirkschumacher: Or move it up to the CPU profile?

run/http_runner.go
```diff
@@ -0,0 +1,267 @@
...
+			wg.Done()
+			return
+		}
+
+		if async {
+			// write the guid to the response.
+			_, err = w.Write([]byte(requestID))
+			if err != nil {
+				handleError(h.httpServer.ErrorLog, async, err, w)
+				return
```
dirkschumacher, line 229: Here is another problem with the `waitGroup`. It does not get released if an error happens.

dirkschumacher (dismissed):
one last call to `wg.Done()` missing it seems

run/generic_runner.go
larsbeck: We can't do that because we want to profile after the run to see memory consumption at that time.

davidrijsman (commented):
In the README of the templates (at least for cluster) it states to invoke
```
nextmv run local . -- -hop.runner.input.path input.json \
  -hop.runner.output.path output.json -hop.solver.limits.duration 10s
```
that should be:

`nextmv run local . -- -runner.input.path input.json -runner.output.path output.json`

dirkschumacher:
@davidrijsman good catch. That applies to all templates

templates/cluster/main.go
```diff
@@ -2,6 +2,7 @@
 package main
```
davidrijsman, line 2: I think the file needs to be:

`// package main holds the implementation of the cluster template.
package main

import (
	"context"
	"log"
	"time"

	"github.com/nextmv-io/sdk/cluster/kmeans"
	"github.com/nextmv-io/sdk/measure"
	"github.com/nextmv-io/sdk/run"
)

type input struct {
	Points        []measure.Point `json:"points"`
	Clusters      int             `json:"clusters"`
	Weights       []int           `json:"weights"`
	MaximumWeight int             `json:"maximum_weight"`
	MaximumPoints int             `json:"maximum_points"`
}

type cluster struct {
	Index    int             `json:"index"`
	Centroid measure.Point   `json:"centroid"`
	Points   []measure.Point `json:"points"`
	Indices  []int           `json:"indices"`
}

// Output is the output of the solver.
type Output struct {
	Clusters          []cluster       `json:"clusters"`
	Feasible          bool            `json:"feasible"`
	Unassigned        []measure.Point `json:"unassigned"`
	UnassignedIndices []int           `json:"unassigned_indices"`
}

func main() {
	err := run.CLI(solver).Run(context.Background())
	if err != nil {
		log.Fatal(err)
	}
}

type ClusterOptions struct {
	Duration time.Duration `json:"duration" default:"10s"`
	Clusters int           `json:"clusters" default:"2"`
}

func solver(input input, opts ClusterOptions) ([]Output, error) {
	// We start by creating a kmeans model.
	model, err := kmeans.NewModel(input.Points, input.Clusters)
	if err != nil {
		return nil, err
	}

	// We set the maximum weight and maximum points for each cluster.
	for _, c := range model.ClusterModels() {
		c.SetMaximumPoints(input.MaximumPoints)

		_, err = c.SetMaximumSumValue(input.MaximumWeight, input.Weights)
		if err != nil {
			return nil, err
		}
	}

	// We create a solver with the model.
	solver, err := kmeans.NewSolver(model)
	if err != nil {
		return nil, err
	}

	// We create the solve options we will use and set the time limit
	// and the measure to use.
	solveOptions := kmeans.NewSolveOptions().
		SetMeasure(measure.EuclideanByPoint()).
		SetMaximumDuration(opts.Duration)

	solution, err := solver.Solve(solveOptions)
	if err != nil {
		panic(err)
	}

	return []Output{format(solution)}, nil
}

// format returns a function to format the solution output.
func format(solution kmeans.Solution) Output {
	o := Output{
		Clusters:          make([]cluster, len(solution.Clusters())),
		Feasible:          solution.Feasible(),
		Unassigned:        solution.Unassigned(),
		UnassignedIndices: solution.UnassignedIndices(),
	}

	for idx, c := range solution.Clusters() {
		o.Clusters[idx] = cluster{
			Index:    idx,
			Centroid: c.Centroid(),
			Points:   c.Points(),
			Indices:  c.Indices(),
		}
	}

	return o
}
`

davidrijsman (approved):

dirkschumacher (approved):

nmisek (approved):

PR #154, opened 2023-01-09
Feature/eng 2136 support clustered routing in platform

davidrijsman:
Introduced a cluster plugin part of sdk
Introduced a cluster template


route/doc.go
```diff
@@ -192,7 +192,7 @@ The following ByPoint implementations are available.
 	HaversineByPoint: Haversine distance between two points
 	TaxicabByPoint:   Taxicab distance between two points
 
-Points may be of any dimension. If the points passed in to any of these
+Unassigned may be of any dimension. If the points passed in to any of these
```
nilsbeck: this is not correct, is it?

templates/cluster/main.go
```diff
@@ -0,0 +1,133 @@
...
+	root := store.New()
+
+	// Add initial solution as nil
+	so := store.NewVar[kmeans.Solution](root, nil)
+
+	i := 0
+	root = root.Generate(func(s store.Store) store.Generator {
+		return store.Lazy(func() bool {
+			// only run one state transition in which we solve the cluster model
+			return i == 0
```
nilsbeck, line 78: The mechanics of store here looks very odd to me. Is there another way to do this?

nilsbeck (dismissed):
looks good. thanks for moving everything to sdk!
just 2 small questions that I was wondering. If we are fine with it the way it is, I certainly am as well.

cluster/kmeans/model.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"github.com/nextmv-io/sdk/measure"
+)
+
+// Model is a model of a k-means clustering problem.
+type Model interface {
+	// ClusterModels returns the cluster models. The cluster models
+	// define the constraints for the clusters.
+	ClusterModels() []ClusterModel
+
```
nmisek, line 12: i would like if this could be called `Cluster` rather than `ClusterModels`.  I found it a bit odd to call .ClusterModels() to access the individual clusters

nmisek: actually - i see why you did it this way.

templates/cluster/main.go
nmisek: hm yes i agree this is a little funny bc it has to be summoned from within store. Am i correct that this will not need to be run from within store once we get general runners out since we can run any algorithm after that point?

nmisek (approved):
Just curious about Nils question on the template but approving

templates/cluster/main.go
davidrijsman: I copy pasted this from the MIP knapsack template to make it a consistent experience in wrapping standalone engine in a store flow.

nilsbeck (approved):

PR #173, opened 2023-02-10
fix bug in clustering

nilsbeck:
# Description

Fixes a bug where cluster size is 0 stops if input has 1 stop and 2 vehicles.


larsbeck (dismissed):

templates/cluster-tsp/main.go
```diff
@@ -119,7 +119,7 @@ func clusterSolution(input input) (kmeans.Solution, error) {
 	}
 
 	for idx := 0; idx < len(input.Vehicles); idx++ {
-		maximumPoints[idx] = len(input.Stops) / len(input.Vehicles)
+		maximumPoints[idx] = (len(input.Stops) / len(input.Vehicles)) + 1
```
dirkschumacher: What is the logic behind `len(input.Stops) / len(input.Vehicles)`? Why does the average number of stops per vehicle equals the maximum points?

I am asking because I am wondering if `s/v + 1` is right here or rather `ceil(float64(s)/float64(v))`?

dirkschumacher (commented):
A comment

templates/cluster-tsp/main.go
nilsbeck: no real requirement. it is mainly to avoid an input problem that could happen in the very rare case of 1s and 2v. We can also use ceil to avoid this problem, I have no opinion.

dirkschumacher: Ah ok, `maximumPoints` is the upper bound on points per cluster. You could also do `(s+v - 1) / v` which is equivalent to `ceil(float64(s)/float64(v))` for integer arithmetic. However you need to ensure that `s >= 0 && v > 0` (was the case before, otherwise there is a panic). In theory `s + v` could overflow, but I think the risk is extremely low here.

Also we can move the computation outside of the loop.

dirkschumacher: That being said, for `v > 0`, `math.Ceil(float64(len(...))/float64(len(...)))` may be more readable.

merschformann: Did you reach a consensus here, or is another opinion required? :sweat_smile:

merschformann: I think I lean towards `Ceil` too, but have no strong opinion. Will rubber stamp to potentially unblock.

merschformann (dismissed):
See comment. :relaxed:

merschformann (approved):

dirkschumacher (approved):

PR #175, opened 2023-02-20
Feature/eng 2309 time depentent measures

nilsbeck:
# Description

This PR gives the option to use DependentMeasures for the routing options TravelTimeMeasures and ValueFunctionMeasures. DependentMeasures contain a set of ByIndex measures of which one is selected by a matching criteria to return the corresponding costs between to points.
Since many use cases need a time dependent measure there is now a time dependent measure client in the route package to conveniently use time dependent measures in a model.


measure/dependent_by_index.go
```diff
@@ -0,0 +1,60 @@
...
+	return &dependentIndexed{
+		cost:          cost,
+		timeDependent: timeDependent,
+	}
+}
+
+// VehicleData holds vehicle specific data, including times by index (ETA, ETD
+// and ETS), a vehicle id, the vehicle's route and the solution value for that
+// vehicle.
+type VehicleData struct {
```
dirkschumacher, line 29: You can reorder the struct to save some bytes:
```go
type VehicleData struct {
	VehicleID  string
	Times      Times
	Route      []int
	Index      int
	RouteValue int
}
```

measure/dependent_by_index.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+// DependentIndexed is a measure uses a custom cost func to calculate parameter
+// dependent costs for connecting to points by index.
+func DependentIndexed(
+	timeDependent bool,
+	cost func(
+		from,
+		to int,
+		data VehicleData,
```
dirkschumacher: Ultimately we would have to benchmark it, but I feel like passing a pointer to that struct instead of a copy may be better here. The struct is about 100 bytes big.

route/option.go
```diff
@@ -181,9 +182,18 @@ func Grouper(groups [][]string) Option {
 // the `Update()` function. In the `Update()` function you can request a
 // `TimeTracker` from the `state` and use it to get access to time information
 // the route.
-func ValueFunctionMeasures(valueFunctionMeasures []ByIndex) Option {
+func ValueFunctionMeasures(
+	valueFunctionMeasures any,
```
dirkschumacher: why not use `[]any` instead of `any`? Then the compiler can at least verify that the input is a slice.

measure/dependent_by_index.go
```diff
@@ -0,0 +1,60 @@
+package measure
+
+import (
+	"encoding/json"
+)
+
+// DependentIndexed is a measure uses a custom cost func to calculate parameter
```
dirkschumacher: Maybe
```
// DependentIndexed is a measure that uses a custom cost function to calculate parameter
// dependent costs for connecting two points by index.
```

route/time_dependent.go
```diff
@@ -0,0 +1,150 @@
...
+	measures        []ByIndexAndTime
+	fallbackMeasure ByIndexAndTime
+	cache           sync.Map
+}
+
+// NewTimeDependentMeasuresClient returns a new NewTimeDependentMeasuresClient
+// which implements a cost function.
+// It takes byIndexAndTime measures, where each measure is given with an endTime
+// (exclusive) up until the measure will be used and a fallback measure.
+func NewTimeDependentMeasuresClient(
```
dirkschumacher: With something that is called `Client` I would expect it makes some kind of connection to a server. Maybe there is a better name?

route/time_dependent.go
```diff
@@ -0,0 +1,150 @@
...
+	measures []ByIndexAndTime,
+	fallback ByIndex,
+	opts ...ClientOption,
+) (TimeDependentMeasuresClient, error) {
+	sort.SliceStable(measures, func(i, j int) bool {
```
dirkschumacher: I think it would be better to create a local copy of the slice and sort that one instead of the parameter.

route/measure.go
```diff
@@ -93,6 +97,19 @@ func Indexed(m ByPoint, points []Point) ByIndex {
 	return measure.Indexed(m, points)
 }
 
+// DependentIndexed is a measure uses a custom cost func to calculate parameter
```
dirkschumacher: Typo here. I think the same text is used at various places.

route/option.go
```diff
@@ -200,9 +210,16 @@ PLEASE NOTE: When defining a custom TravelTimeMeasure, this measure must not
 account for any service times. To account for services times please use the
 Services option.
 */
-func TravelTimeMeasures(timeMeasures []ByIndex) Option {
+func TravelTimeMeasures(timeMeasures any) Option {
```
dirkschumacher: AFAICT we use `any` here to not break the api, right? Why don't we introduce a new option called that is type safe:
* `TravelTimeMeasures([]ByIndex) Option`
* `TravelTimeDependentMeasures([]DependentByIndex) Option`
Same applies to `ValueFunctionMeasures`. That would leave everything backwards compatible and we would not loose any compiler safety.

route/time_dependent.go
```diff
@@ -0,0 +1,150 @@
...
+// calculate costs.
+type TimeDependentMeasuresClient interface {
+	Cost() func(from, to int, data measure.VehicleData) float64
+	DependentByIndex() measure.DependentByIndex
+}
+
+type client struct {
+	measures        []ByIndexAndTime
+	fallbackMeasure ByIndexAndTime
+	cache           sync.Map
```
dirkschumacher: What motivated you to introduce a cache vs. maintaining a static mapping from time to measure?

dirkschumacher (commented):
Just some early comments. Have to drop, but will come back later 😅

route/time_dependent.go
nilsbeck: basically a full pre-cache was a wish @davidrijsman had from past experiences. to leave it open how to use cache, the full pre-cache is an option that can be passed. otherwise the mapping will be created on the fly and stored for future usages.

route/option.go
nilsbeck: yes, that works and makes sense, I guess

measure/dependent_by_index.go
nilsbeck: fixed

route/measure.go
nilsbeck: fixed

route/time_dependent.go
nilsbeck: done

route/time_dependent.go
nilsbeck: I think we can just remove the client. Makes it also easier to use. 👍

route/option.go
nilsbeck: with the change to introduce a new option for dependent value functions/travel time functions, this becomes obsolete.

measure/dependent_by_index.go
nilsbeck: done

measure/dependent_by_index.go
nilsbeck: changed it.

route/time_dependent.go
dirkschumacher: How many time dependent measures do we expect a user to have? In orders of magnitude? 10, 100, 1000?

dirkschumacher: My gut says there is no need for a cache. If I understand it correctly, given a startTime , we will choose the first measure whose endTime is greater than startTime.

We know that the runtime overhead is always constant in the number of stops/vehicles and only depends on the number of measures.

For a smaller number of measures we can simply loop through the sorted list of measures and return as soon as the above relation is true: `O(n), where n is the number of measures`, but very simple code. Since the list is sorted, we can also use a binary search that brings us to `O(log n)` run time probably. The overhead in code complexity makes only sense if we expect a very large number of measures.

I have the feeling these two approaches would outperform the cache based approach, both in terms of runtime and code complexity. Only for very large inputs I would expect the cache to perform better. But haven't tested it - maybe I am totally wrong here.

route/time_dependent.go
```diff
@@ -0,0 +1,148 @@
...
+	EndTime int
+}
+
+// ClientOption can pass options to be used with a TimeDependentMeasure client.
+type ClientOption func(*client)
+
+// TimeDependentMeasure is an interface to handle time dependent
+// measures. It implements a Cost function that takes time into account to
+// calculate costs.
+type TimeDependentMeasure interface {
```
dirkschumacher: Why is that interface exposed? Do we expect users to create their own implementations? It seems this is only used internally.

nilsbeck: it just became unused with "client" change. so it will be removed.

route/time_dependent.go
nilsbeck: I guess there won't be many measures - I could be wrong though.
That's by I think it's okay to loop over the sorted measures to find the correct one and then store it in the map for future reference and just have a look up.
For other cases if many measures are used customers can use the option and create cache if they see fit?

I am not seeing a downside here?

dirkschumacher: If we don't need a cache, then I think the logic to create a time dependent measure can be greatly simplified I think.

davidrijsman: Don't forget this is in the deepest inner loop you are adding an O(logn) lookup instead of O(1). Not sure if this is adding complexity or just potentially a lot of memory consumption (map[time]measure). But if too much complexity I would say try first simple (more time expensive) solution

dirkschumacher: Until we have some good benchmarks setup for this case, we could go with the O(1) approach under the hood.

nilsbeck: done

route/time_dependent.go
```diff
@@ -0,0 +1,118 @@
...
+// NewTimeDependentMeasure which a slice of ByIndexAndTime.
+type ByIndexAndTime struct {
+	Measure ByIndex
+	EndTime int
+}
+
+type client struct {
+	measures        []ByIndexAndTime
+	fallbackMeasure ByIndexAndTime
+	cache           map[int]ByIndexAndTime
```
dirkschumacher, line 21: Instead of mapping to a reference to a `ByIndexAndTime` you could also map to the measure index (`map[int]int`). This is a bit more memory efficient I would expect.

route/time_dependent.go
```diff
@@ -0,0 +1,118 @@
...
+		from,
+		to,
+		measure.EndTime,
+		newCosts,
+		partialFactor-newPartialFactor,
+	)
+}
+
+func cacheTimes(startTime int, c *client) {
+	counter := 0
```
dirkschumacher: what is the purpose of `counter`?

route/time_dependent.go
nilsbeck: I know, we discussed it. but when I implemented it it felt more natural to use a map because the index is not starting from 0 and needs a shift from the start time. I can do that but it feels weird.

route/time_dependent.go
nilsbeck: left over. removed. thanks

route/time_dependent.go
dirkschumacher: I am not suggesting to use a slice, but a map from int to int instead of int to ByIndexAndTime.

dirkschumacher (approved):

nmisek (approved):

merschformann (approved):

PR #180, opened 2023-02-23
Addresses already created tags

merschformann:
# Description

Tags are already present when the release workflow is invoked. This PR addresses it.


merschformann:
Will do a fast merge, since this has zero impact on vital parts / is non-breaking.

PR #215, opened 2023-03-08
feature/eng-2551-provide-route_distance-on-every-vehicle

nilsbeck:
# Description

This PR adds a new field route_distance to the solution's output.

### DIFF
--- a/measure/interface.go
+++ b/measure/interface.go
@@ -41,7 +41,7 @@ func IsTriangular(m any) bool {
 // Times holds the estimated time of arrival (ETA), estimated time of when
 // service starts (ETS) and estimated time of departure (ETD).
 type Times struct {
-	EstimatedArrival      []int `json:"estimated_arrival"`
-	EstimatedServiceStart []int `json:"estimated_service_start"`
-	EstimatedDeparture    []int `json:"estimated_departure"`
+	EstimatedArrival      []int `json:"estimated_arrival,omitempty"`
+	EstimatedServiceStart []int `json:"estimated_service_start,omitempty"`
+	EstimatedDeparture    []int `json:"estimated_departure,omitempty"`
 }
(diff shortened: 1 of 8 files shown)


measure/interface.go
```diff
@@ -41,7 +41,7 @@ func IsTriangular(m any) bool {
 // Times holds the estimated time of arrival (ETA), estimated time of when
 // service starts (ETS) and estimated time of departure (ETD).
 type Times struct {
-	EstimatedArrival      []int `json:"estimated_arrival"`
-	EstimatedServiceStart []int `json:"estimated_service_start"`
-	EstimatedDeparture    []int `json:"estimated_departure"`
+	EstimatedArrival      []int `json:"estimated_arrival,omitempty"`
```
dirkschumacher, line 44: Why `omitempty` now?

route/router.go
```diff
@@ -97,6 +97,7 @@ type PlannedVehicle struct {
 	ID            string        `json:"id"`
 	Route         []PlannedStop `json:"route"`
 	RouteDuration int           `json:"route_duration"`
+	RouteDistance int           `json:"route_distance"`
```
dirkschumacher, line 100: What is the unit? Maybe Call it `RouteDistance<Unit>` to make it clear what the distance is? Same then applies to the json name

measure/interface.go
nilsbeck: because we not always have ETA etc (if no shift is given) and in this case the output has fields like `"estimated_arrival": <nil>` In legacy router we omitted them also (not a good reason, I know 😄 )
but I can revert that. Just felt weird to have it in there when there is no information.

route/router.go
nilsbeck: mh, I like it short. we always work with seconds and meters - the standard SI units. And if we make that change we would need to change `RouteDuration` as well to state that it is in seconds.

larsbeck (approved):

nmisek (approved):

PR #216, opened 2023-03-09
feature/eng-2533-bug-with-measure-matrix-sdk-functions

nilsbeck:
# Description

This PR makes the HERE client more robust when a Point of (0,0) is passed. Now those points are interpreted correctly as points that do not need to be routed and a cost of 0 is returned when going to or coming from such a point.

## Changes

- `here/measure.go`

### DIFF
--- a/route/here/measure.go
+++ b/route/here/measure.go
@@ -95,12 +95,14 @@ func NewClient(apiKey string, opts ...ClientOption) Client {
 }
 
 func cleanPoints(points []route.Point) []route.Point {
-	for i, p := range points {
+	cleanPoints := make([]route.Point, len(points))
+	copy(cleanPoints, points)
+	for i, p := range cleanPoints {
 		if len(p) == 2 && p[0] == 0 && p[1] == 0 {
-			points[i] = route.Point{}
+			cleanPoints[i] = route.Point{}
 		}
 	}
-	return points
+	return cleanPoints
 }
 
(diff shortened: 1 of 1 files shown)


route/here/measure.go
```diff
@@ -94,13 +94,23 @@ func NewClient(apiKey string, opts ...ClientOption) Client {
 	return c
 }
 
+func cleanPoints(points []route.Point) []route.Point {
+	for i, p := range points {
+		if len(p) == 2 && p[0] == 0 && p[1] == 0 {
+			points[i] = route.Point{}
```
dirkschumacher: Why don't we remove the points then vs. setting it to the zero-value struct?

route/here/measure.go
```diff
@@ -94,13 +94,23 @@ func NewClient(apiKey string, opts ...ClientOption) Client {
 	return c
 }
 
+func cleanPoints(points []route.Point) []route.Point {
```
dirkschumacher, line 97: This modifies in place. I think it would be better if we made a copy and clean that.

route/here/measure.go
nilsbeck: because then we change the matrix size from MxM to a smaller matrix and this matrix then cannot be used in router because it requires the MxM

route/here/measure.go
nilsbeck: yes, made a copy first

nmisek (approved):

dirkschumacher (approved):

PR #217, opened 2023-03-09
feature/eng-2577-create-a-new-template-for-unique-point

nilsbeck:
# Description

This PR adds a new template to show how to use a unique matrix to be passed in an input file.

## Changes

- `main.go` for new template
- `readme.md` for new template
- `input.json` for new template
- test for new template


nmisek (changes requested):
Mentioned in other PR but I think would be good to have vehicle starts/ends in here

templates/unique-matrix-measure/main.go
```diff
@@ -0,0 +1,178 @@
...
+	for stopIndex1 := range fullMatrix {
+		// Create an row filled with 0.
+		fullMatrix[stopIndex1] = make([]float64, matrixSize)
+
+		// Get first reference index.
+		refIndex1 := joinedStops[stopIndex1].Reference
+		for stopIndex2 := range fullMatrix[stopIndex1] {
+			// Get second reference index and look up costs.
+			refIndex2 := joinedStops[stopIndex2].Reference
+			if refIndex1 != -1 && refIndex2 != -1 {
```
larsbeck, line 75: If they are both `-1` then the cost in `fullMatrix` will be `0`. Is that what we want here?

nilsbeck: yes, that would mean that we are going from/to a start or end that is not given in the input. Those arcs should be cost 0.

larsbeck (dismissed):

nmisek (dismissed):

nmisek (dismissed):

larsbeck (approved):

nmisek (approved):

PR #224, opened 2023-03-29
Change all fields to be pointers

sebastian-quintero:
# Description

Harmonizes the input by making all non-mandatory fields pointers. This is needed for the new logic that processes defaults.

### DIFF
--- a/nextroute/schema/input.go
+++ b/nextroute/schema/input.go
@@ -10,14 +10,14 @@ import (
 // Input is the default input schema for nextroute.
 type Input struct {
 	// Options contains arbitrary options specified by the user.
-	Options        *any                 `json:"options,omitempty"`
-	Defaults       *Defaults            `json:"defaults,omitempty"`
-	Vehicles       []Vehicle            `json:"vehicles,omitempty"`
-	Stops          []Stop               `json:"stops,omitempty"`
-	StopGroups     [][]string           `json:"stop_groups,omitempty"`
-	AlternateStops []Stop               `json:"alternate_stops,omitempty"`
-	DurationGroups []route.ServiceGroup `json:"duration_groups,omitempty"`
-	DurationMatrix [][]float64          `json:"duration_matrix,omitempty"`
+	Options        *any                  `json:"options,omitempty"`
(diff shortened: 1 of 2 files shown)


davidrijsman (approved):

nextroute/schema/input.go
```diff
@@ -10,14 +10,14 @@ import (
...
 	// Options contains arbitrary options specified by the user.
-	Options        *any                 `json:"options,omitempty"`
-	Defaults       *Defaults            `json:"defaults,omitempty"`
-	Vehicles       []Vehicle            `json:"vehicles,omitempty"`
-	Stops          []Stop               `json:"stops,omitempty"`
-	StopGroups     [][]string           `json:"stop_groups,omitempty"`
-	AlternateStops []Stop               `json:"alternate_stops,omitempty"`
-	DurationGroups []route.ServiceGroup `json:"duration_groups,omitempty"`
-	DurationMatrix [][]float64          `json:"duration_matrix,omitempty"`
+	Options        *any                  `json:"options,omitempty"`
```
nilsbeck: Do we actually need `Options` and if  we do, shouldn't this be of type `NextRouteOptions` (currently still found in `platform/nextapp/main.go`)?

nilsbeck: other than that: a pointer to an interface is not really needed but probably does not hurt either

nilsbeck (commented):
just a question that we should make a decision upon: how do we make the `NextRouteOptions`visible to the user - if we want to integrate them. (I believe we do)

nextroute/schema/input.go
sebastian-quintero: Completely right on the pointer to interface thing, my bad. Other than that, @larsbeck what should we do about these options and of what type should they be?

larsbeck: They are currently there for @merschformann to test in cloud. Ultimately we want a way to pass options (and then they will be `NextRouteOptions` in this case) to cloud, but we aren't there yet. Let's leave them there for now as we are still experimenting and remove them once we merge.

PR #225, opened 2023-03-29
Feature/eng 2701 create a router 10 hidden template and

nilsbeck:
# Description

Adds a hidden template for informs23.


templates/routing-informs/README.md
```diff
@@ -0,0 +1,38 @@
+# Nextmv INFORMS Analytics 2023 routing template
+
+The INFORMS template sets up a vehicle routing problem (VRP) solver that is
+compatible with our [cloud](https://docs.nextmv.io/cloud/get-started) input
+files. In addition to almost being a drop-in replacement for the cloud endpoint
+it also demonstrates some of the more advanced router options.
+
+The most important files created are `main.go` and several data input files for
+different uses cases: fleet management, delivery, distribution and sourcing. By
```
nmisek: Can we remove the note about different use cases? That’s I think copy paste error from cloud routing template.

nmisek: Also I think we need some edits to this since you brought the code into a single file

nmisek (changes requested):

nmisek (dismissed):

templates/routing-informs/encoder.go
```diff
@@ -0,0 +1,244 @@
...
+		}
+		return nil
+	default:
+		return errors.New("invalid duration")
+	}
+}
+
+type statisticsOut struct {
+	Schema string `json:"schema"`
+	Run    struct {
```
dirkschumacher: Is `statisticsOut.Run.Time` set somewhere? Otherwise this is always 0. You could also leave out `Run`, as it is optional.

nilsbeck: this is the runtime that was initally given, right? then, I can kill it.
And the schema should be `v1`?

nmisek (approved):

mamclaughlin (approved):

PR #226, opened 2023-03-31
ENG-2704 Enable creation of invalid locations

sebastian-quintero:
# Description

Regarding the use of `Location`:
- Create the `NewInvalidLocation` function to enable the creation of invalid locations. 
- Modify the `NewLocation` function to also return an error.
- Change `Location` to be an interface, to avoid having two ways to instantiate a new `Location`. Now, only `NewLocation` or `NewInvalidLocation` may be used.
- Change signatures of functions that rely on `NewLocation` and `NewInvalidLocation` to return an error as well.

Regarding building a new `Model`:
- Creates the function `BuildModel` to return a ready-to-go `Model`.
- Deletes the factory logic.
- Enables the `SetName` method on the `ModelExpression` interface to allow custom naming of default expressions. This is used to add the vehicle's ID when using a TravelDurationExpression.

Regarding `common`:
- Added the `UniqueDefined` func to allow unique comparisons on interface types.

### DIFF
--- a/nextroute/common/fast_haversine.go
+++ b/nextroute/common/fast_haversine.go
@@ -38,7 +38,7 @@ func wrap(deg float64) float64 {
 
 // Distance returns the distance between two locations in meters.
 func (f FastHaversine) Distance(from, to Location) float64 {
-	if !from.valid || !to.valid {
+	if !from.IsValid() || !to.IsValid() {
 		return 0.0
 	}
 	dx := wrap(from.Longitude()-to.Longitude()) * f.kx
--- a/nextroute/common/fast_haversine.go
+++ b/nextroute/common/fast_haversine.go
@@ -1,6 +1,9 @@
 package common
 
-import "math"
+import (
+	"fmt"
+	"math"
+)
 
 // NewFastHaversine returns a new FastHaversine.
 func NewFastHaversine(lat float64) FastHaversine {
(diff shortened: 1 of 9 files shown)


nextroute/common/fast_haversine.go
```diff
@@ -38,7 +38,7 @@ func wrap(deg float64) float64 {
 
 // Distance returns the distance between two locations in meters.
 func (f FastHaversine) Distance(from, to Location) float64 {
```
davidrijsman: Why does Haversine throw an error and FastHaversine return zero/

nextroute/common/location.go
```diff
@@ -44,46 +50,60 @@ func (l Locations) Unique() Locations {
 
 // Centroid returns the centroid of the locations. If locations is empty, the
 // centroid will be (0, 0).
-func (l Locations) Centroid() Location {
+func (l Locations) Centroid() (Location, error) {
```
davidrijsman, line 53: Should we return an invalid location if locations is empty, centroid of nothing is not in the middle of the ocean (0,0)

davidrijsman (changes requested):

nextroute/common/location.go
nilsbeck: also, can the Locations have an invalidLocation among them? should this be considered when calculating the centroid? Meaning: only add up lat and lon for locations that are valid locations?

nilsbeck (approved):

nextroute/common/fast_haversine.go
sebastian-quintero: 🤷🏻 Haven't been thorough with the code yet 😅 . I'll return 0 and an error.

nextroute/common/location.go
sebastian-quintero: Ok so now it returns an invalid location is the length of locations is 0 and it validates if a location is invalid, in which case an error is returned (as opposed to just ignoring it).

PR #228, opened 2023-04-04
ENG-2761 Create the shift constraint

sebastian-quintero:
# Description

Enables the `SetEpoch` method on the model to be able to customize the model's epoch.


sebastian-quintero:
@davidrijsman and @larsbeck pointed out that we do not want to allow a user to change the epoch because of the potential harm it can do to expressions.

PR #230, opened 2023-04-05
feature/eng-2721-create-a-distance-limit-option

nilsbeck:
# Description

This PR introduces a distance limit modifier.

## Changes

- adds distance matrix to input
- adds ignore option for new option
- removes obsolete code

### DIFF
--- a/nextroute/model.go
+++ b/nextroute/model.go
@@ -21,14 +21,15 @@ func NewModel() (Model, error) {
 // features to the model, such as constraints and objectives. On the other
 // hand, [NewModel] creates an empty vehicle routing model which must be built
 // from the ground up.
-func BuildModel(i schema.Input, m ModelOptions, opts ...Option) (Model, error) {
+func BuildModel(i schema.Input, m ModelOptions) (Model, error) {
 	connect.Connect(con, &buildModel)
-	return buildModel(i, m, opts...)
+	return buildModel(i, m)
 }
 
 // ModelOptions represents options for a model.
 type ModelOptions struct {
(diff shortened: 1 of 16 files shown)


nextroute/model_options.go
```diff
@@ -1,8 +0,0 @@
-package nextroute
-
-import (
-	"github.com/nextmv-io/sdk/nextroute/schema"
-)
-
-// An Option configures a model.
-type Option func(Model, schema.Input) error
```
larsbeck: Why did this get removed? We'll need them for Options that add code, such as custom value function.

larsbeck (changes requested):

nextroute/model_options.go
nilsbeck: see comment in platform

larsbeck (approved):

sebastian-quintero (approved):

nextroute/model.go
```diff
@@ -28,7 +28,8 @@ func BuildModel(i schema.Input, m ModelOptions, opts ...Option) (Model, error) {
 
 // ModelOptions represents options for a model.
 type ModelOptions struct {
-	IgnoreUnplanned bool `json:"ignore_unplanned"  usage:"ignore unplanned penalties"`
+	IgnoreUnplanned          bool `json:"ignore_unplanned"  usage:"ignore unplanned penalties"`
+	IgnoreRouteDistanceLimit bool `json:"ignore_route_distance_limit"  usage:"ignore route distance limit"`
```
davidrijsman: These options are they being used, I don't see how?

davidrijsman (approved):
Just a question on modeloptions but should not block approval

PR #235, opened 2023-04-06
Feature/eng 2799 cluster/inwardness option

davidrijsman:
The inwardness constraint forces stops to be added to vehices whose centroid is closest to the stops added. Enabling this constraint creates compact, visually appealing, routes.


larsbeck (approved):

merschformann (approved):
:rocket:

PR #238, opened 2023-04-06
ENG-2709 Enable the max_duration constraint

sebastian-quintero:
# Description

Adds a couple of fields to the `ModelOptions` and removes the `RouteDistance` from the basic output as we do not have the capability of calculating distance yet.

### DIFF
--- a/nextroute/schema/output_basic.go
+++ b/nextroute/schema/output_basic.go
@@ -18,7 +18,6 @@ type VehicleOutput struct {
 	ID            string       `json:"id"`
 	Route         []StopOutput `json:"route"`
 	RouteDuration int          `json:"route_duration"`
-	RouteDistance int          `json:"route_distance"`
 }
 
 // StopOutput adds information to the input stop.
--- a/nextroute/factory/model.go
+++ b/nextroute/factory/model.go
@@ -24,14 +24,16 @@ func NewModel(
 
 // ModelOptions represents options for a model.
 type ModelOptions struct {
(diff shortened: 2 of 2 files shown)


nextroute/schema/output_basic.go
```diff
@@ -18,7 +18,6 @@ type VehicleOutput struct {
 	ID            string       `json:"id"`
 	Route         []StopOutput `json:"route"`
 	RouteDuration int          `json:"route_duration"`
-	RouteDistance int          `json:"route_distance"`
```
larsbeck, line 21: I believe that was added already. The necessary distanceExpression came in with this PR https://github.com/nextmv-io/platform/pull/296

sebastian-quintero: If you do not use that constraint (for example if it is disabled via options or it is not specified in the input) then that expression is not added anywhere. @davidrijsman and I discussed this morning and we have to find a good way to report distance even if it is not used in constraints or objectives. There is a ticket for that already: [ENG-2800](https://linear.app/nextmv/issue/ENG-2800/add-distance-reporting-to-the-output-in-the-basic-formatter)

sebastian-quintero: Also, the API currently does not have any method that allows you to get the cumulative distance for a vehicle or vehicle type.

larsbeck: Sounds good, thanks for clarification

davidrijsman (approved):

PR #243, opened 2023-04-12
Feature/eng 2710 create a router 20 hidden template

nilsbeck:
# Description

This PR adds a hidden template for router2.0 for informs23.


sebastian-quintero:
@nilsbeck I think we can close this in favor of #255, what do you think?

nilsbeck:
> @nilsbeck I think we can close this in favor of #255, what do you think?

Yes, this was created for informs when we decided to not use it after all.

PR #248, opened 2023-04-13
Proposed reorganization for nextroute package

sebastian-quintero:
# Description

This PR is a proposal, so it might as well be rejected. It proposes a slight reorganization of the `nextroute` package, mainly focused around deleting the `common` and `schema` packages therein.

- This [entry in the official Go Blog](https://go.dev/blog/package-names) explicitly states:
> Avoid meaningless package names. Packages named util, common, or misc provide clients with no sense of what the package contains
- Dave Cheney [says](https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common):
> Avoid package names like base, util, or common

I agree with these opinions (after all, they are _just_ opinions), that's why I think it is better to provide our users with clear declarations. Here are some examples that I find are an improvement in syntax:
- `common.NewLocation` 👉🏻 `nextroute.NewLocation`.
- `common.DefensiveCopy` 👉🏻 `slice.DefensiveCopy`
- `schema.Input` 👉🏻 `nextroute.Input`.

Here is a good quote:
> A package’s name provides context for its contents, making it easier for clients to understand what the package is for and how to use it.

That is why `common` (and `schema`) ultimately gets broken into:
- `slice`
- `statistics`
- `alias`
- `nextroute` 👉🏻 a bunch of stuff gets moved back to the main package.

# Changes

Here is a breakdown of what changed and my reasoning for it.
- Everything under `nextroute/common/utils.go` that was related to wrangling slices was moved to its own `nextroute/slice` package because ultimately all the functions provide features around working with slices.
- Everything under `nextroute/schema` got moved to `nextroute`. I believe that having a package exclusively for the schema adds no value and I think that `nextroute.Input` is more straightforward than `schema.Input`.
- Everything related to location, speed and distance got moved out of `nextroute/common` and into `nextroute`. I don't see how `common.Location` or `common.NewSpeed` has a better meaning than it being part of the `nextroute` package itself, so now stuff like `nextroute.NewLocation` gives me a `nextroute.Location`. I think location, speed and distance stuff being part of the main package is better because after all that is the place where it is being used and it is not "common" among other packages.
- `statistics.go` got moved to `nextroute/statistics` pkg and `alias.go` got moved to `nextroute/alias` package.
- To comply with the style that is used throughout the `nextroute` package, where the user has factory functions (`NewFoo`) that return an interface and the implementation is hidden across the plugin barrier, I proposed we do the same with `Location`, `Speed`, `SpeedUnit`. Now the implementation is hidden and the style is the same as other factory methods.
- Of course, all references were correctly changed.


davidrijsman (changes requested):
I understand that there may be differing opinions on the best way to organize the code. However, I have carefully considered and implemented a consistent approach that prioritizes readability, maintainability, and correctness. While there may be other valid perspectives, I believe it's important to stick with the current structure for now and focus on completing the functionality with those goals in mind.

PR #253, opened 2023-04-14
Informs workshop script feedback

cowanwalls:
# Description

Addressed feedback from informs workshop

## Changes

- `templates/routing-informs/part1.sh`
    - Updated file permissions to add execute privileges
    - Removed lowercase casting through `tr` and just expect failure when user does not enter initials as explained and expected
- `templates/routing-informs/part2.sh`
    - Updated file permissions to add execute privileges
    - Addressed line length consistency
- `templates/routing-informs/part3.sh`
    - Updated file permissions to add execute privileges
    - Addressed line length consistency


sebastian-quintero (dismissed):

templates/routing-informs/part1.sh
chriskeyser, line 1: `Then, set the API key you used to configure to the following value:
export NEXTMV_API_KEY=<your key>`

This is not needed as we check at the start of the script that the api key is set and will never get to this point if it isn't.

chriskeyser (commented):
see below comment.

chriskeyser (approved):

PR #259, opened 2023-04-21
feature/eng-2794-add-vehicletypedistanceexpression

nilsbeck:
# Description

This PR adds a new `VehicleExpression` that pulls together the previously existing `VehicleTypeExpression` and the new `VehicleTypeDistanceExpression`. The `VehicleExpression` is used in a `CustomMaxConstraint` now to allow limitations on both, `VehicleTypeExpression` and `VehicleTypeDistanceExpression`. The latter is now used for limiting distances and makes sure the units for distance are being used uniformly by internally using `Dinstance` and `DistanceUnit.`


nextroute/model_expression_custom.go
```diff
@@ -51,6 +51,19 @@ func NewVehicleTypeExpression(
...
 }
 
+// NewVehicleTypeDistanceExpression returns a NewVehicleTypeDistanceExpression
+// whose value is based on the vehicle type in Value method and expresses a
+// distance. Expression values are calculated by calling the expression's Value
+// method which takes a vehicle type, a from stop and a to stop as arguments.
+func NewVehicleTypeDistanceExpression(
+	name string,
+	defaultValue common.Distance,
+	defaultUnit common.DistanceUnit,
```
davidrijsman: Not sure why we need defaultUnit, a distance is a combination of a value and unit. So I think we already have the unit in the distance.

nextroute/model_expression_custom.go
```diff
@@ -124,10 +137,16 @@ type StopExpression interface {
 	)
 }
 
+// VehicleExpression is an expression that implements a DefaultExpression and is
```
davidrijsman: We do not require a VehicleExpression, we can have the VehicleTypeDistanceExpression be a VehicleTypeExpression.

nextroute/model_expression_custom.go
```diff
@@ -124,10 +137,16 @@ type StopExpression interface {
...
+// used be encapsuled by VehicleTypeExpression and VehicleTypeDistanceExpression.
+type VehicleExpression interface {
+	DefaultExpression
+}
+
 // VehicleTypeExpression is an expression that has a value for each vehicle
 // type.
 type VehicleTypeExpression interface {
-	DefaultExpression
+	VehicleExpression
```
davidrijsman: This can be a default expression

nextroute/model_expression_custom.go
```diff
@@ -136,6 +155,19 @@ type VehicleTypeExpression interface {
 	)
 }
 
+// VehicleTypeDistanceExpression is a distance expression that has a value for
+// each vehicle type.
+type VehicleTypeDistanceExpression interface {
+	VehicleExpression
```
davidrijsman: This can be a VehicleTypeExpression (something that is dependent on the type of vehicle) or cleaner it can be a DefaultExpression

nextroute/model_expression_custom.go
```diff
@@ -136,6 +155,19 @@ type VehicleTypeExpression interface {
...
+
+	// SetValue sets the distance value of the expression for the given
```
davidrijsman: This can stay the same if we decide by making VehicleTypeDistanceExpression a DefaultExpression implementation and otherwise you have to change it to SetDistance (no method overloading allowed)

nextroute/engine.go
```diff
@@ -139,6 +139,11 @@ var (
 		string,
 		float64,
 	) VehicleTypeExpression
+	newVehicleTypeDistanceExpression func(
+		string,
+		common.Distance,
+		common.DistanceUnit,
```
davidrijsman: We do not need the distance-unit (already part of distance)

davidrijsman (dismissed):
Some naming and redundant interfaces

sebastian-quintero (dismissed):
Pending David's comments are processed

larsbeck (approved):

PR #261, opened 2023-04-25
Adds maximum wait constraint

merschformann:
# Description

Adds constraint for limiting wait time (time not moving / serving stops).

## Changes

- Adds wait time constraint


nextroute/schema/input.go
```diff
@@ -67,6 +68,7 @@ type Vehicle struct {
 	MaxStops                *int            `json:"max_stops,omitempty"`
 	MaxDistance             *int            `json:"max_distance,omitempty"`
 	MaxDuration             *int            `json:"max_duration,omitempty"`
+	MaxWait                 *int            `json:"max_wait,omitempty"`
```
sebastian-quintero: This is proposing that we add a max wait to the vehicle... Would that be confusing given that we have always defined max wait on the stop?

sebastian-quintero (changes requested):
Requesting changes only because this is a "breaking" functionality change from previous routers.

nextroute/schema/input.go
merschformann: Yep, that is correct. This is an incremental change and should be backwards compatible, which is why I am fine with it. :joy_cat: 
It was a comment by @davidrijsman , since he expects customers to more likely require a constraint on accumulated wait (i.e., don't make one driver wait all their shift) instead of single stop wait (don't wait more than X at a single stop - which is potentially a strange constraint). Anyway, I don't have a strong opinion here, but think it's useful and doesn't harm.

merschformann: Oh, and maybe there was something confusing here: the max_wait was already defined on the stop. It was just not implemented under the hood yet. So, you won't see it as a diff in my PR.

sebastian-quintero: Ok, more documentation for me 😅 .

sebastian-quintero (approved):

nextroute/schema/input.go
merschformann: Haha, actually I wanted to ping you already: do you have a docs project somewhere? Feel free to add an issue for me to add docs for this feature here. I just don't know yet where the docs will go for this (maybe you have them already though - a bit out of the loop on docs here :grimacing: ).

PR #269, opened 2023-05-01
Feature/eng 2918 balance stops objective

larsbeck:


sebastian-quintero:
nextroute moved to another repo

PR #271, opened 2023-05-02
Fix comment style and rename RunStart

larsbeck:
# Changes

- Some renaming of symbols to prevent "stuttering"
- Make comments start with lower case letters and not end in punctuation

### DIFF
--- a/run/encoder.go
+++ b/run/encoder.go
@@ -93,22 +93,14 @@ func (g *genericEncoder[Solution, Options]) Encode(
 		for solution := range solutions {
 			m.Solutions = append(m.Solutions, solution)
 		}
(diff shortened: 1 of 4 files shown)


run/encoder.go
```diff
@@ -93,22 +93,14 @@ func (g *genericEncoder[Solution, Options]) Encode(
 		for solution := range solutions {
 			m.Solutions = append(m.Solutions, solution)
 		}
-		if err = g.encoder.Encode(ioWriter, m); err != nil {
-			return err
-		}
-
-		return nil
+		return g.encoder.Encode(ioWriter, m)
```
merschformann, line 96: :heart:

merschformann (approved):
:rocket:

PR #283, opened 2023-05-15
Aligning input and output formats

larsbeck:
# Description

This PR aligns input and output field names. It also adds statistics and therefore resolves #ENG-3078. It furthermore adds the time to solution in the statistics section and therefore resolves #ENG-3072. It also resolves #ENG-3057 by adding progression to the statistics section. This PR adds a context to `run.CLI` and `run.HTTP` - this is needed to determine when we started a run, which we need to compute the duration of a run as part of statistics.

Resolves ENG-3020, ENG-3078, ENG-3072, ENG-3057, ENG-3087


davidrijsman (changes requested):
How do we envision users to add custom data to the output and/or how can I locally use my verbose output?

larsbeck:
> How do we envision users to add custom data to the output and/or how can I locally use my verbose output?

There are `CustomData` fields on the `schema.Output` struct that you can use.

davidrijsman (approved):

PR #286, opened 2023-05-16
Return dates in the timezone of the user

larsbeck:
# Description

Users of `nextroute` provide `times` in their own timezones. This PR ensures that we output them in the same timezone as they came in.

### DIFF
--- a/nextroute/solution_format.go
+++ b/nextroute/solution_format.go
@@ -144,9 +144,10 @@ func setTimes(
 	stopOutput schema.PlannedStopOutput,
 	hasUserDefinedStartTime bool,
 ) schema.PlannedStopOutput {
-	arrival := solutionStop.Arrival()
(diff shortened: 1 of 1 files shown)


nextroute/solution_format.go
```diff
@@ -144,9 +144,12 @@ func setTimes(
 	stopOutput schema.PlannedStopOutput,
 	hasUserDefinedStartTime bool,
 ) schema.PlannedStopOutput {
-	arrival := solutionStop.Arrival()
-	departure := solutionStop.End()
-	service := solutionStop.Start()
+	// we need to access the timezone via the vehicle of the model
```
larsbeck, line 147: The vehicle that is accessible via the `solutionStop` directly has lost the timezone information.

sebastian-quintero (approved):

PR #291, opened 2023-05-19
Add RangeMap to allow deterministic ranging over maps

larsbeck:
# Description

Ranging over maps in Go deliberately happens in random order. This is ok in some cases, but in others it is not. In our case, whenever we have a break criteria in the loop ranging over a map, this will randomly influence the runtime. This PR introduces a helper function that can be used to safely range over maps.


dirkschumacher (approved):
Interesting, wondering what the performance overhead is. But maybe it doesn't matter. Some (benchmark) tests would have been nice as well.

PR #298, opened 2023-05-25
address gosec issues

wenfeng-nextmv:
Address issues found by gosec and remove all rule exclusions from security scan action.


nextroute/common/utils.go
```diff
@@ -262,20 +269,43 @@ func RandomElementIndices[T any](
...
 	if source == nil {
+		// using unsafe math/rand is just a unlikely fallback, we can ignore the gosec issue here
+		// G404 (CWE-338): Use of weak random number generator (math/rand instead of crypto/rand)
+		/* #nosec */
 		source = rand.New(rand.NewSource(time.Now().UnixNano()))
 	}
 
 	for {
-		index := source.Intn(size)
+		index := RandIntnWithFallback(source, size)
```
larsbeck: We are using `RandomIndex` as part of our algorithm. Pseudo-random number generators are fine in this context. We do not worry about them not being cryptographically safe. What is more important is that they are fast. Unless there is something I am missing, I would like us to keep using `rand.Rand`.

route/osrm/client.go
```diff
@@ -128,7 +128,7 @@ func (c *client) get(uri string) (data []byte, err error) {
 	var key string
 
 	if c.useCache {
-		key = fmt.Sprintf("%x", sha1.Sum([]byte(uri)))
+		key = fmt.Sprintf("%x", sha256.Sum256([]byte(uri)))
```
larsbeck: Same here. Speed matters the most in this case and this is not a security relevant operation. We are creating a key for a cache. With that said `sha1` is probably also not the best (fastest) choice, but I do not see the need to use `sha256`.

larsbeck (dismissed):

larsbeck:
I think we should keep the `gosec` exceptions in here and not change the code as these are algorithmic internals that do rely on speed, but not on cryptographic safety. Maybe I am missing how an attacker might use the fact that we are using non-cryptographically safe random stream to make random decisions or break ties in the algorithm. If so, could you give an example of how that would be done?

nextroute/common/utils.go
wenfeng-nextmv: Revert to use math/rand and add #nosec after discussion.

route/osrm/client.go
wenfeng-nextmv: Revert to sha1 and add #nosec after discussion.

davidrijsman (approved):

templates/measure-matrix/main.go
```diff
@@ -109,6 +109,13 @@ func main() {
 
 	// Write the output.
 	file, _ := json.MarshalIndent(out, "", " ")
+
+	// [gosec issue] G306 (CWE-276): Expect WriteFile permissions to be 0600 or less
+	// macOS default file permission is 0644 -rw-r--r--
+	// Linux default file permission is 0660 -rw-rw-r--
+	// This file we created contains no secret or credential. 0644 permission should not have any security risk.
+	// If a user adds sensitive content in this file later, it is the user's responsibility to protect it.
+	/* #nosec */
```
merschformann: This file is part of a template that will be user facing (generate via: `nextmv sdk init -t measure-matrix`). I think we should avoid having such internal comments in there. Maybe we can add the template dir to the exclusions?

.github/workflows/gosec.yml
```diff
@@ -13,7 +13,7 @@ jobs:
         with:
           # For now, exclude some rules that we currently have issues with, so that PR merges won't be blocked.
           # After we address those issues, we shall remove the excluded rules.
-          args: "-no-fail -exclude=G306,G401,G404,G505 -fmt csv -out results.csv -stdout -verbose=text -exclude-generated -exclude-dir=template.go -exclude-dir=sample_code -exclude-dir=node_modules/aws-cdk -exclude-dir=cdk.out ./..."
+          args: "-no-fail -fmt csv -out results.csv -stdout -verbose=text -exclude-generated -exclude-dir=template.go -exclude-dir=sample_code -exclude-dir=node_modules/aws-cdk -exclude-dir=cdk.out ./..."
```
merschformann: I think there are some copy / paste exclusions here that are not relevant for the sdk repo, right? :thinking: 
I.e., `sample_code`, `node_modules/aws-cdk` & `cdk.out`?

route/osrm/client.go
```diff
@@ -128,6 +131,11 @@ func (c *client) get(uri string) (data []byte, err error) {
 	var key string
 
 	if c.useCache {
+		// sha1 is used to shorten the key for faster cache lookup than directly using the lenthy uri as key.
+		// The chance of hash colision is extremely low for sha1.
+		// The cache is local to the user, which won't become a secrity threat even when key colides.
```
merschformann:
```suggestion
		// The cache is local to the user, which won't become a security threat even when key collides.
```
Just a small typo. :relaxed:

merschformann (commented):
Thanks for the work! :hugs: 
Just a small remark about the comment in the `templates/...` file. Maybe we should exclude that whole directory? :thinking:

.github/workflows/gosec.yml
wenfeng-nextmv: You are right. These are for cloud/ and not needed here.

route/osrm/client.go
wenfeng-nextmv: Fixed. Good eyes!

templates/measure-matrix/main.go
wenfeng-nextmv: I am removing most of the comment but leaving `#nosec`. For now I am a little hesitated to exclude the entire `templates/` dir from gosec scan. After all, we are responsible to provide good, secure template code samples for users to get started safely.

cowanwalls (approved):

PR #299, opened 2023-05-26
Renames hard window in input.json

nmisek:
Renames `hard_window` to `start_time_window`


sebastian-quintero (approved):
@nmisek Good call. I think this is why we need to add some JSON validation to the Runner. So this stuff does not fly under the radar.

PR #304, opened 2023-06-05
Change `initialization_cost` to `activation_penalty`

sebastian-quintero:
# Description

I believe this is the last part of the input that needed to be updated to be consistent. All other objectives that are penalties are named `penalty` so this PR does the same for the former initialization cost.


merschformann (approved):
:rocket:

PR #309, opened 2023-06-19
Feature/eng 3208 confusing api to cast newdurationexpression

davidrijsman:
* NewDurationExpression inline with NewDistanceExpression
* Introduced NewScaledDurationExpression

### DIFF
--- a/nextroute/common/duration.go
+++ b/nextroute/common/duration.go
@@ -1,0 +1,69 @@
+package common
+
+import (
+	"fmt"
+	"time"
+)
+
+// DurationUnit is the unit of duration.
+type DurationUnit int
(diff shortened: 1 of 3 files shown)


nextroute/common/duration.go
```diff
@@ -0,0 +1,69 @@
+package common
+
+import (
+	"fmt"
+	"time"
+)
+
+// DurationUnit is the unit of duration.
+type DurationUnit int
```
sebastian-quintero, line 9: Why is all of this necessary and we don't simply use Go built-in's `time.Duration`?

sebastian-quintero (approved):

PR #314, opened 2023-06-27
Added options for validating nextroute input

davidrijsman:
# Description

Validation has been extende and options for factory input have been introduced:

* `Options.Validate.Disable.StartTime`, allow to disable the start time validation which requires a start time on a vehicle if start time windows on stops have been defined
* `Options.Validate.Enable.Matrix`, enable validation of values in distance and duration matrices
* `Options.Validate.Enable.MatrixAsymmetryTolerance`, allows to define the percentage symmetric cells can deviate from each other if matrices are validated


sebastian-quintero (approved):

nilsbeck (approved):

merschformann (approved):
:rocket:

PR #316, opened 2023-06-28
feature/eng-2988-cloud-routing-app-parity-support-duration-groups

nilsbeck:
# Description

This PR adds duration groups to nextroute.

### DIFF
--- a/nextroute/factory/model.go
+++ b/nextroute/factory/model.go
@@ -64,6 +64,7 @@ type Options struct {
 	Properties struct {
 		Disable struct {
 			Durations       bool `json:"durations" usage:"ignore the durations of stops"`
+			GroupDurations  bool `json:"group_durations" usage:"ignore the group durations of stops"`
 			InitialSolution bool `json:"initial_solution" usage:"ignore the initial solution"`
 		} `json:"disable"`
 	} `json:"properties"`
--- a/nextroute/factory/model.go
+++ b/nextroute/factory/model.go
@@ -64,7 +64,7 @@ type Options struct {
 	Properties struct {
 		Disable struct {
 			Durations       bool `json:"durations" usage:"ignore the durations of stops"`
-			GroupDurations  bool `json:"group_durations" usage:"ignore the group durations of stops"`
+			DurationGroups  bool `json:"duration_groups" usage:"ignore the durations groups of stops"`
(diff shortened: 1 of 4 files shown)


nextroute/factory/model.go
```diff
@@ -64,6 +64,7 @@ type Options struct {
 	Properties struct {
 		Disable struct {
 			Durations       bool `json:"durations" usage:"ignore the durations of stops"`
+			GroupDurations  bool `json:"group_durations" usage:"ignore the group durations of stops"`
```
davidrijsman: It is DurationGroups

nextroute/schema/input.go
```diff
@@ -101,3 +102,10 @@ type Location struct {
 	Lon float64 `json:"lon"`
 	Lat float64 `json:"lat"`
 }
+
+// DurationGroup represents a group of stops that get additional process time
```
davidrijsman: I believe we stay away from calling it process time, just use _duration_

davidrijsman (changes requested):

nextroute/schema/input.go
```diff
@@ -101,3 +102,10 @@ type Location struct {
...
+// (duration) whenever a stop of the group is approached for the first time.
+type DurationGroup struct {
+	Group    []string `json:"group,omitempty"`
```
sebastian-quintero, line 109: Maybe we call this `Stops`? The name of the struct already alludes to the fact that this is a group...

sebastian-quintero (commented):
Would like for you to process comments before directly approving 🙏🏻 .

nextroute/factory/model.go
nilsbeck: fixed

nextroute/schema/input.go
nilsbeck: fixed

nextroute/schema/input.go
nilsbeck: If we do that we need to name the field in the input json also stops. I feel that this is confusing, since it is a duration group and they are not duration stops. So I think, while may seem repetitive in the code, it is still the better option when we look at the user side of things.

larsbeck (approved):

davidrijsman (approved):

PR #319, opened 2023-07-07
Improves feedback on failed initial solution

merschformann:
# Description

Improves the error message when failing to create the initial solution due to constraint violation.


nextroute/model_constraint_maximum.go
```diff
@@ -20,11 +20,13 @@ type MaximumConstraint interface {
 }
 
 // NewMaximumConstraint creates a new maximum constraint. The constraint
-// needs to be added to the model to be taken into account.
+// needs to be added to the model to be taken into account. Furthermore, a name
+// identifying the constraint needs to be provided.
 func NewMaximumConstraint(
 	expression ModelExpression,
 	maximum VehicleTypeExpression,
+	name string,
```
merschformann: @davidrijsman I wasn't fully sure about the order of the parameters here. I default to name last and probably picked it up back when working with Gurobi (see [here](https://www.gurobi.com/documentation/9.5/refman/java_model_addconstr.html)). Let me know, if we roll differently. :relaxed:

sebastian-quintero: We should avoid introducing further breaking changes. I would create a new function: `NewMaximumConstraintWithName` or something like that.

sebastian-quintero (changes requested):
I would not introduce any more breaking changes, unless absolutely necessary.

nextroute/model_constraint_maximum.go
davidrijsman: In general I like to apply the rule that the factory method takes the minimum set of arguments that are required to define a functioning instance of whatever it is creating. If possible it should be possible to create a working instance by just using the factory function. Name is definitely not required for a functioning instance of the constraint, therefor I would create a setter on constraint for name (we do this for `ModelStop` and `ModelVehicle` where we have setters for the ID). Also this would not break the existing API.

davidrijsman (changes requested):
See comments

nextroute/model_constraint_maximum.go
merschformann: That is also a breaking change for custom constraints (they need to implement another func to fulfill the interface). Also, I am not sure whether a somewhat anonymous max constraint will ever be a good call. :shrug: If it's used in different ways that is. 
But I am ok with a different way. There are other constraints that could suffer from the same problem tbh. Maybe we just don't name them differently and accept the problem with not being able to distinguish capacity / distance limit / potential custom constraints basing on it. :thinking: 
I am really undecided, but lean slightly to always require a name when creating one (would need to be extended to the others too).
Let me know what you think.

merschformann: P.S. I am not a fan of having of another func that does the same just to add a string argument, I think. That feels verbose to me and specifying a name for a constraint doesn't seem like a big deal. But yeah, it's one way out without breaking for sure.
Regarding breaking: how many people use this yet and are we out of beta? We said we will be changing the API quite a bit in the beginning as we did not have a proper design phase around it. Did we pass that point? Just checking to make sure how I should go about long term benefit vs. breaking change. It would certainly change my opinion here too. Maybe not distinguishing the constraints in more detail would be good enough even. :shrug:

davidrijsman: * "they need to implement another func to fulfill the interface", they do not have to, if Name() is not implemented you can revert to stringer, if that is not implemented you can revert to type of the constraint which should be sufficient given it is a custom constraint.
* "I am not a fan of having of another func that does the same just to add a string argument", I am with you, if argument required we should add it but it is not required
* and yes we can still break API's but I don't think we have to because of the reasons already mentioned

merschformann: I think there is a misunderstanding here. Can you clarify where we would add `Name()`?
If we add it to the `ModelConstraint` interface (see [here](https://github.com/nextmv-io/sdk/blob/develop/nextroute/model_constraint.go#L105-L116)), it needs to be implemented by every (custom / user) constraint.
But I suppose you are talking about something different. Are you suggesting to add another interface? I guess that would work. One with a setter and getter for the name. It could be a generic one too, I suppose. :thinking:

davidrijsman: Yes, I was thinking of adding a Namer (or Identifier to be consistent with `ModelStop` and `ModelVehicle`) interface with two methods and check if the constraint implements that
```
type Namer interface {
   Name() string 
   SetName(string)
}
```
```
type Identifier interface {
   ID() string 
   SetID(string)
}
```

davidrijsman (approved):

sebastian-quintero (approved):

PR #320, opened 2023-07-07
Fix schemas and inputs

larsbeck:


templates/mip-knapsack/input.json
merschformann, line 1: lol, where did the volume come from? I guess there was a multi-dimensional knapsack at some point. :sweat_smile:

route/router.go
```diff
@@ -117,8 +117,8 @@ type Stop struct {
 
 // TimeWindow represents a time window for a shift of a vehicle or a stop.
 type TimeWindow struct {
-	Start time.Time `json:"start"`
-	End   time.Time `json:"end"`
+	Start time.Time `json:"start,omitempty"`
+	End   time.Time `json:"end,omitempty"`
```
merschformann, line 121: Out of curiosity: why is `omitempty` required on input structs? Isn't that only relevant for `Marshal` and not so much `Unmarshal`.

merschformann (approved):
:rocket:

sebastian-quintero (approved):

route/router.go
larsbeck: The json schema validation library marks fields with omitempty  as optional. And as you correctly say, _we_ only use it on outputs for marshaling via the json package in the standard library. So this should not interfere.

templates/mip-knapsack/input.json
larsbeck: It was there for a demo so that people only had to add fields to the struct and then add the new constraint.

route/router.go
merschformann: Ah, makes sense. Thanks for clarifying.

PR #327, opened 2023-07-20
ENG-3348 Add default statistics

sebastian-quintero:
# Description

Add default statistics to the `mip` templates and `nextroute` template. The idea is to have an example of how to pass in custom statistics that is:
1. Backwards compatible.
2. Not breaking the API.

With this, even if people forget to add a custom something, experiments will look more interesting as they offer custom statistics from the get go.


mip/solution_format.go
```diff
@@ -0,0 +1,63 @@
...
+		status = "suboptimal"
+	case solution.IsInfeasible():
+		status = "infeasible"
+	}
+
+	return CustomResult{
+		Status:  status,
+		Columns: len(model.Vars()),
+		Rows:    len(model.Constraints()),
+	}
```
merschformann: Good call, I like it.
I doubt that we have non-zeros, right? Also, we probably have no access to solve stats like cuts added, pre-solve time, pre-solve reduction, etc.. So, ignore this comment. We should get back here, if we support them some day though. :relaxed:

mip/options.go
```diff
@@ -0,0 +1,8 @@
+package mip
+
+import "time"
+
+// Limits on the MIP solver, such as how long it can run for.
+type Limits struct {
+	Duration time.Duration `json:"duration" usage:"maximum duration of the solver. A duration limit of 0 is treated as infinity" default:"30s"`
```
merschformann, line 9: Sometimes I wish we would comment these fields so that intellisense (or whatever we call it) picks up on it. Even though it's the input structs, even I have used them directly before.
However, just leaving a thought. We're not doing it anywhere else, so, no need to start here and now. :relaxed:

nextroute/solution_format.go
```diff
@@ -21,6 +21,19 @@ type FormatOptions struct {
...
+	UsedVehicles      int `json:"used_vehicles,omitempty"`
+	UnplannedStops    int `json:"unplanned_stops,omitempty"`
+	MaxTravelDuration int `json:"max_travel_duration,omitempty"`
+	MaxDuration       int `json:"max_duration,omitempty"`
+	MinTravelDuration int `json:"min_travel_duration,omitempty"`
+	MinDuration       int `json:"min_duration,omitempty"`
+	MaxStopsInRoute   int `json:"max_stops_in_route,omitempty"`
+	MinStopsInRoute   int `json:"min_stops_in_route,omitempty"`
+}
+
```
merschformann: Should this be part of the `schema` maybe? :thinking: (no strong opinion, just thinking it may sit well with the output struct - or near it)
And same comment about field comments / docs. Again, we don't even do it for the core input struct, so, no need to do it here.

nextroute/solution_format.go
```diff
@@ -130,6 +143,58 @@ func Format(
...
+		if stops > maxStops {
+			maxStops = stops
+		}
+		if stops < minStops {
+			minStops = stops
+		}
+	}
+
+	return CustomStatistics{
+		UsedVehicles:      vehicleCount,
```
merschformann: Another one: we call the related option `ActivationPenalty`. Should we call it `ActivatedVehicles` instead to align the language? :shrug:

nextroute/solution_format.go
```diff
@@ -130,6 +143,58 @@ func Format(
...
+		UnplannedStops:    solution.UnPlannedPlanUnits().Size(),
+		MaxTravelDuration: maxTravelDuration,
+		MaxDuration:       maxDuration,
+		MinTravelDuration: minTravelDuration,
+		MinDuration:       minDuration,
+		MaxStopsInRoute:   maxStops,
+		MinStopsInRoute:   minStops,
```
merschformann: It is probably too much, but I personally would love to also have these as median. Or, have mean + stddev? :shrug: 
Again, just thoughts. Feel free to ignore.

templates/mip-incentive-allocation/main.go
```diff
@@ -25,62 +23,44 @@ func main() {
...
+// Input of the problem.
+type input struct {
+	// Users for the problem including name and a cost/effect pair per
+	// available incentive.
+	Users []struct {
+		ID string `json:"id"`
+		// Incentives that can be applied to the user.
+		Incentives []struct {
+			ID string `json:"id"`
+			// Positive effect of the incentive.
```
merschformann, line 40: Huh, here we use some field comments. :relaxed: 
Field comments would also avoid the constant reformatting of the struct when something in it changes. Sorry, I will let go of the thought now. :joy_cat:

mip/solution_format.go
```diff
@@ -0,0 +1,63 @@
...
+	value := statistics.Float64(solverSolution.ObjectiveValue())
+	output.Statistics.Result.Value = &value
+
+	output.Solutions = []any{solution}
+
+	return output
+}
+
+// DefaultStatistics creates default custom statistics for a given solution.
+func DefaultStatistics(model Model, solution Solution) CustomResult {
```
merschformann: I find the naming here a bit confusing. Can't we rather just add these statistics on the format itself? I would like to always have them, I think. 
If not, I would prefer something like this, I think:
```go
func AddExtendedStatistics(model Model, solution Solution, output schema.Output) {}
```
Or something like that. 
Alternatively, we can rename this to `mip.DefaultStatisticsResult` or similar. 
The reason I am a bit confused by it is that the Statistics (and DefaultStatistics suggest that) is more than the result part. There is also the progression part which we typically add in nextroute for example.

Let me know, if I don't make any sense. :sweat_smile:

nextroute/solution_format.go
```diff
@@ -130,6 +143,58 @@ func Format(
 	return output
 }
 
+// DefaultStatistics creates default custom statistics for a given solution.
+func DefaultStatistics(solution Solution) CustomStatistics {
```
merschformann: Same comment here about the naming / way we add custom statistics (see above). :relaxed:

merschformann (approved):
Looks great and a big improvement imho.
I just have a small concern regarding the way we add the statistics and whether we always add them by default. I don't want to be blocking though. Maybe someone else can add their thoughts as well? cc @larsbeck @davidrijsman

mip/solution_format.go
sebastian-quintero: Yeah, those are all the stats that made sense to me in the short term.

mip/options.go
sebastian-quintero: I can comment, no problem!

nextroute/solution_format.go
sebastian-quintero: Sure, I'll move to schema. And I can comment the fields.

nextroute/solution_format.go
sebastian-quintero: Yep, no problem.

nextroute/solution_format.go
sebastian-quintero: The goal of this is to show some default stats so that the experiments come up nicely... I wouldn't want to be completely exhaustive here, for it is just a sample.

templates/mip-incentive-allocation/main.go
sebastian-quintero: Will try to apply field comments always.

mip/solution_format.go
sebastian-quintero: The reason why I added the statistics as a separate function is to make it super clear in the template that there are these things called statistics. If it is buried inside the `Format` function, then it all happens automagically, so I wanted to shed visibility there. I think it is nicer for people to simply plug out the default function and use their own function to apply stats to the output. That is why it is separate.

On another note, I'll change the naming. Thanks.

nextroute/solution_format.go
sebastian-quintero: Yep, see above.

nextroute/solution_format.go
merschformann: Yeah, it's hard to draw a line. This is fine. :relaxed:

nextroute/schema/output.go
```diff
@@ -54,3 +54,32 @@ type ObjectiveOutput struct {
...
+	// MinDuration is the minimum duration of a vehicle (including waiting
+	// times) in the solution, excluding vehicles that were not used.
+	MinDuration int `json:"min_duration,omitempty"`
+	// MaxStopsInRoute is the maximum number of stops in a vehicle's route in
+	// the solution. The start and end stops of the vehicle are not considered.
+	MaxStopsInVehicle int `json:"max_stops_in_vehicle,omitempty"`
+	// MinStopsInRoute is the minimum number of stops in a vehicle's route in
+	// the solution. The start and end stops of the vehicle are not considered.
+	MinStopsInVehicle int `json:"min_stops_in_vehicle,omitempty"`
+}
```
merschformann, line 85: Thank you for the comments! :hugs:

merschformann (approved):
:heart:

PR #330, opened 2023-07-21
Fix IOProducer returning an empty error

larsbeck:
# Description

The IOProducer returned an empty error where it should have returned the actual error that caused the termination of the runner.


merschformann (approved):
:rocket:

PR #331, opened 2023-07-24
Adds multi windows constraint support

merschformann:
# Description

Adds multi-window support. See internal PR for further details on the implementation.

## Changes

- Adds multi-window support


nextroute/model_stop.go
```diff
@@ -43,17 +43,23 @@ type ModelStop interface {
...
 	PlanUnit() ModelPlanUnit
 
 	// SetEarliestStart sets the earliest start time of the stop.
-	SetEarliestStart(time time.Time)
+	// The start of a stop will either be later than t or within one of
+	// the intervals defined in intervals. The intervals are not allowed
+	// to overlap and must be strictly increasing. t must be larger than
+	// the end of the last interval. If the intervals are empty, the
+	// earliest start time is set to t.
+	SetEarliestStart(intervals [][2]time.Time, t time.Time) error
```
larsbeck: Why do we need a separate value `t`? Should the intervals be a variadic argument so that the is not a breaking change?

merschformann: @davidrijsman and me debated the different ways of doing this for a bit. We came to the conclusion that just submitting one window would cause a quite substantial rework of the architecture, as the window end is handled by a constraint while the earliest start (and further windows are functionally the same) is handled here.
Since we have windows in the input / factory anyway and this is the lower technical way of instantiating things, we decided that this will work.
We can re-think this, but it will cascade into the general architecture too, I think.

merschformann: I totally agree that this is confusing though. :100:

davidrijsman: Hmm not sure what is confusing or not confusing and the impact on the architecture.... Just make sure the user can do anything we want it to do and make sure itcan not do what we do not want it to do. So if a user invokes this, it should provide at least one earliest start (with end infinity) and it potentially can add more intervals, given that the intervals need to be ordered in time, the current API makes perfect sense. Maybe we need to document it different.

sebastian-quintero (commented):
Is any customer using this API directly? I'm against the breaking changes, I would've done as variadic arguments.

larsbeck (approved):

PR #332, opened 2023-07-26
Feature/eng 2800 add distance reporting to the output in the basic formatter

davidrijsman:
Distance has been added to basic formatter of solutions.

### DIFF
--- a/templates/nextroute/main.go
+++ b/templates/nextroute/main.go
@@ -47,8 +47,8 @@ func solver(
 	}
 	last := solutions.Last()
 
-	output := nextroute.Format(ctx, options, solver, last)
-	output.Statistics.Result.Custom = nextroute.DefaultCustomResultStatistics(last)
+	output := factory.Format(ctx, options, solver, last)
+	output.Statistics.Result.Custom = factory.DefaultCustomResultStatistics(last)
 
 	return output, nil
 }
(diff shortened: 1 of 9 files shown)


templates/nextroute/main.go
```diff
@@ -47,8 +47,8 @@ func solver(
 	}
 	last := solutions.Last()
 
-	output := nextroute.Format(ctx, options, solver, last)
-	output.Statistics.Result.Custom = nextroute.DefaultCustomResultStatistics(last)
+	output := factory.Format(ctx, options, solver, last)
```
sebastian-quintero, line 50: This is a breaking change and we should not introduce it. I don't see the benefit of shielding the format function, it is just statistics calculation and it actually makes sense to keep it public to reference when someone is implementing a custom formatter.

sebastian-quintero (dismissed):
I would rather not introduce this breaking change because if someone uses a new version, their current `nextroute` code stops working. Other than that, introducing distance is a great idea.

templates/nextroute/main.go
davidrijsman: We would only be able to not shield the nextroute.Format as the factory.Format is using non exported structures to calculate the distance. Is that what you would like to export, I exported the nextroute.DefaultCustomResultStatistics for the exact reason you are stating, as an example how to enhance/change the formatter?

sebastian-quintero (commented):
@davidrijsman and I chatted this thoroughly and the reasoning is that `nextroute` should not know about `schema` or `factory`. This is a fair point and I agree. However, I'm still against the breaking change. I don't want to keep breaking customers, even if it's just an import change. So, agree to disagree 😸 .

sebastian-quintero:
Feel free to remove my request for changes.

larsbeck (approved):

PR #335, opened 2023-08-03
fixed up err handling and no start/stop

chriskeyser:
# Summary

This PR fixes a few issues with OSRM and differentiates user caused errors:
* [deferred: https://linear.app/nextmv/issue/ENG-3549/address-osrm-issues] We were always adding start and ends for points in `BuildMatrixRequestPoints`. This added some extraneous `[0,0]` points if start and ends were not defined in the input.
* If an error occurred within the `client.Table` in one of the goroutines during the parallel requests to OSRM, then the function returned immediately upon any error occurrence. Goroutines may have remained running in the background. Returning caused a deferred close of the signaling channel, and the remaining goroutines then sent on a closed channel, causing a panic. Updated the logic to wait for all goroutines to complete, collect all errors, and return them to the client.
* Added differentiation between system caused errors, and user caused errors. We assume a 400 returned from OSRM is a result of bad user input.

# Testing

* Added unit testing for additional logic. 
* Tested with integration test for pre-processing handling.


types/user_error.go
sebastian-quintero, line 1: I would prefer to keep `sdk` as nimble as possible. Defining a whole new package which is just used inside `route/osrm` seems a bit on a overkill. I would just add this `UserError` inside `route/osrm`, as unexported.

sebastian-quintero (approved):

types/user_error.go
chriskeyser: How are we differentiating user caused errors today in sdk, or aren't we? This has been a problem in general that we've had - the inability to differentiate between an error that is something internal to us and an error that is a result of user input failure. How do we do this with nextroute for example? I'm happy to follow that model.

chriskeyser:
Think there is an issue where default vehicle start/end wasn't being looked at, fixing that. The input had a default start/end

Still should look thru this with someone tomorrow to make sure I understand assumptions though. There is a bug that I'm working to fix but I still don't yet have it all figured out. In particular I don't think the logic would ever work if there are not both start and end.

The error I saw when I didn't add default vehicles:


```
'--------------------------------------------------------------------------------\n' +
  'This software is provided by Nextmv.\n' +
  'The current Nextmv token expires in 14 minutes.\n' +
  '\n' +
  '© 2019-2023 nextmv.io inc. All rights reserved.\n' +
  '\n' +
  '    (\\/)     (\\/)     (\\/)     (\\/)     (\\/)     (\\/)     (\\/)     (\\/)     (\\/)\n' +
  '    (^^)     (^^)     (^^)     (^^)     (^^)     (^^)     (^^)     (^^)     (^^)\n' +
  '   o( O)    o( O)    o( O)    o( O)    o( O)    o( O)    o( O)    o( O)    o( O)\n' +
  '--------------------------------------------------------------------------------\n' +
  'panic: runtime error: index out of range [55] with length 55\n' +
  '\n' +
  'goroutine 6 [running]:\n' +
  'github.com/nextmv-io/sdk/measure.matrix.Cost(...)\n' +
  '\tgithub.com/nextmv-io/sdk@v0.25.0/measure/matrix.go:14\n' +
  'github.com/nextmv-io/platform/core/plugins/nextroute.(*measureByIndexExpression).Value(0x40001a3d10, {0xffff678f1710?, 0x4000149098?}, {0xffff67cf0b28?, 0x400007fe00?}, {0xffff67cf0b28, 0x400007fe80})\n' +
  '\tgithub.com/nextmv-io/platform/core/plugins/nextroute/model_expression_measure_byindex.go:54 +0x74\n' +
  'github.com/nextmv-io/platform/core/plugins/nextroute.(*scaledDurationExpressionImpl).Value(0x40001a3d40, {0xffff67cf06c8?, 0x40001093b0?}, {0xffff67cf0b28?, 0x400007fe00?}, {0xffff67cf0b28?, 0x400007fe80?})\n' +
  '\tgithub.com/nextmv-io/platform/core/plugins/nextroute/model_expression_duration.go:149 +0x48\n' +
  'github.com/nextmv-io/platform/core/plugins/nextroute.(*timeIndependentDurationExpressionImpl).ValueAtValue(0x352bc0?, 0x40002b6ad8?, {0xffff67cf06c8?, 0x40001093b0?}, {0xffff67cf0b28?, 0x400007fe00?}, {0xffff67cf0b28?, 0x400007fe80?})\n' +
  '\tgithub.com/nextmv-io/platform/core/plugins/nextroute/model_expression_time_dependent.go:513 +0x40\n' +
  'github.com/nextmv-io/platform/core/plugins/nextroute.(*vehicleTypeImpl).TemporalValues(0x40001093b0, 0x41d85b0e9c000000, {0xffff67cf0b28, 0x400007fe00}, {0xffff67cf0b28, 0x400007fe80})\n' +
  '\tgithub.com/nextmv-io/platform/core/plugins/nextroute/model_vehicle_type.go:56 +0xa4\n' +
  'github.com/nextmv-io/platform/core/plugins/nextroute.(*solutionImpl).isFeasible(0x400014cf00, 0x37, 0x1)\n' +
  '\tgithub.com/nextmv-io/platform/core/plugins/nextroute/solution.go:842 +0x440\n' +
  'github.com/nextmv-io/platform/core/plugins/nextroute.(*solutionImpl).isNotFeasible(...)\n' +
  '\tgithub.com/nextmv-io/platform/core/plugins/nextroute/solution.go:750\n' +
  'github.com/nextmv-io/platform/core/plugins/nextroute.(*solutionImpl).newVehicle(0x400014cf00, {0xffff67cf0650, 0x4000109420})\n' +
  '\tgithub.com/nextmv-io/platform/core/plugins/nextroute/solution.go:626 +0xbfc\n' +
  'github.com/nextmv-io/platform/core/plugins/nextroute.NewSolution({0xffff67cf2e20?, 0x400011d860})\n' +
  '\tgithub.com/nextmv-io/platform/core/plugins/nextroute/solution.go:162 +0xbe4\n' +
  'github.com/nextmv-io/platform/core/plugins/nextroute.(*parallelSolverImpl).Solve(0x40000146e0, {0x43f7c8, 0x400001b9e0}, {0x7fffffffffffffff, 0x6fc23ac00, 0xffffffffffffffff, 0x6, 0x0}, {0x0?, 0x0, ...})\n' +
  '\tgithub.com/nextmv-io/platform/core/plugins/nextroute/solver.go:219 +0x200\n' +
  'main.solver({_, _}, {{0x0, 0x0}, 0x400016a8e0, 0x0, 0x400000ec78, 0x40001ad698, {0x4000035c00, 0x5, ...}, ...}, ...)\n' +
  '\texample.com/your_project/nextroute/main.go:44 +0xfc\n' +
  'github.com/nextmv-io/sdk/run.CLI[...].func1({{0x0, 0x0}, 0x400016a8e0, 0x0, 0x400000ec78, 0x40001ad698, {0x4000035c00, 0x5, 0x6}, {0x40001f4000, ...}, ...}, ...)\n' +
  '\tgithub.com/nextmv-io/sdk@v0.25.0/run/simple_runner.go:19 +0xa4\n' +
  'github.com/nextmv-io/sdk/run.(*genericRunner[...]).Run.func2()\n' +
  '\tgithub.com/nextmv-io/sdk@v0.25.0/run/generic_runner.go:158 +0xf4\n' +
  'created by github.com/nextmv-io/sdk/run.(*genericRunner[...]).Run\n' +
  '\tgithub.com/nextmv-io/sdk@v0.25.0/run/generic_runner.go'
```

types/user_error.go
sebastian-quintero: I think right now we are not differentiating between user errors and engine errors. When an error happens it is simply surfaced. Something that even happens in `nextroute` is that in some places functions panic, instead of returning an error.

sebastian-quintero:
@chriskeyser this is the error with that file.

```bash
[~/Downloads]$ cat 2d923f0b-3ec8-41ff-86d9-af5b9ffc1aef.streets.json | jq ".stops | length"
55
[~/Downloads]$ cat 2d923f0b-3ec8-41ff-86d9-af5b9ffc1aef.streets.json | jq ".vehicles | length"
5
[~/Downloads]$ cat 2d923f0b-3ec8-41ff-86d9-af5b9ffc1aef.streets.json | jq ".duration_matrix | length"
55
[~/Downloads]$ cat 2d923f0b-3ec8-41ff-86d9-af5b9ffc1aef.streets.json | jq ".duration_matrix[] | length"
55
55
55
...
```

You have 55 stops and 5 vehicles. [Per the docs](https://docs.nextmv.io/docs/vehicle-routing/features/duration-matrix), the matrix should be:

```
n + 2m
```

Where:
- `n`: the number of stops.
- `m`: the number of vehicles.

Right now your matrix is 55x55 but it should be 65x65.

This means that vehicles' starts and ends should always be included, even if they are not present. They should be included as invalid locations.

sebastian-quintero:
@chriskeyser When there are no vehicles' starts and/or ends, you populate the missing data with 0... This is actually described in the docs:
> Vehicle locations are omitted, given that the duration of going to/from them is zero.

sebastian-quintero:
@chriskeyser The fix (for me) would be for the `OSRM` client to produce a matrix with `0`s standing-in for missing vehicles' starts/ends locations.... I think that is why this line originally works:

```go
points := make([]Point, count+2*vehiclesCount)
```

Because empty points are generated for that matrix...

chriskeyser:
@sebastian-quintero @larsbeck @davidrijsman 

> Unfortunately this doesn't appear to work when you don't have an OSRM service that is the whole world. I talked this thru with David and chatted with Lars there are a few issues:

* Implementation doesn't work unless whole worlds since you have to generate routes to `0,0` in osrm
* When you have whole world, we are doing work on the OSRM service that could be avoided by generating routes to `0,0`

We really should collapse a number of things we are doing with the APIs I believe after speaking with David which is beyond what I can get done for this PR into a single higher level function (we could still leave the other functions exposed):
* We should have a higher level function for customers to call.
* We should only submit valid points for the routes to OSRM
* We should build out the matrices for distance and duration after OSRM runs, plugging points that aren't defined (ie today `0,0` into the matrices at the appropriate points for vehicles as `0,0`) into the matrices. This isn't rocket science but will be non-trivial IMHO.

For now David's suggesting as a work around is to just insert the last stop if the vehicle start or end isn't defined. I'll take that avenue for now and I'll update this PR to just include the improved error handling (please call out if you have concerns with that approach)

types/user_error.go
chriskeyser: That's not a great model. How do we differentiate things we need to worry about (ie the service has an issue) vs. things the customer did incorrectly? We need to differentiate especially for the subscription service. I'd like to understand the case for panicking better. It just seems like something that we shouldn't to do me.

sebastian-quintero: I agree 100% with you.

sebastian-quintero (approved):

PR #336, opened 2023-08-04
propagate error rather than sending a new error when occurs for osrm

chriskeyser:
# Summary

We were sending a new error rather than propagating the error from `DistanceMatrix` which hides the user input error raised from below. This fixes that issue.


merschformann (approved):
:rocket:

PR #338, opened 2023-08-09
Sanity check of a model

davidrijsman:
It is now possible to sanity check a model. Changing the solver content with this will report a sanity check as solution.

```
func solver(ctx context.Context,
	input schema.Input,
	options options,
) (runSchema.Output, error) {
	model, err := factory.NewModel(input, options.Model)
	if err != nil {
		return runSchema.Output{}, err
	}
	output, err := nextroute.NewModelSanityCheck(model).Check(
		options.Solve.Duration,
		sdkNextRoute.SanityCheckMedium,
	)
	if err != nil {
		return runSchema.Output{}, err
	}
	
	return nextroute.Format(
		ctx,
		options.Format,
		nil,
		func(_ sdkNextRoute.Solution) any {
			return output
		},
		nil,
	), nil
}
```


nextroute/model_sanity_check.go
```diff
@@ -0,0 +1,163 @@
...
+)
+
+// SanityCheckDepth is the depth of the sanity check.
+type SanityCheckDepth int
+
+const (
+	// SanityCheckLow checks if there is at least one move per plan unit.
+	SanityCheckLow SanityCheckDepth = iota
+	// SanityCheckMedium checks same as SanityCheckLow.
+	SanityCheckMedium
```
merschformann, line 16: Is this meant as a placeholder for the future then?

nextroute/model_sanity_check.go
```diff
@@ -0,0 +1,163 @@
...
+	Check(duration time.Duration, depth SanityCheckDepth) (SanityCheckOutput, error)
+}
+
+// SanityCheckOutput is the output of the sanity check.
+type SanityCheckOutput struct {
+	// Remark is the remark of the sanity check. It can be "ok", "timeout" or
+	// anything else that should explain itself.
+	Remark string `json:"remark"`
+	// Error is the error raised during the sanity check. It can be empty.
+	Error string `json:"error"`
```
merschformann, line 37: Can we clarify a bit more whether this would contain a summary error or is an internal error of the check operation that wouldn't be expected? Without having had a look at the details I would assume that this is a high-level error message about some sanity checks firing, as the internal check error is returned separately from the output.

nextroute/model_sanity_check.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// SanityCheckInitialSolution is the sanity check of the initial solution. The
+// initial solution is the solution created by the solver with the initial stops
+// that are fixed. The non-fixed stops are not planned.
+type SanityCheckInitialSolution struct {
+	// Feasible is true if the initial solution is feasible.
+	Feasible bool `json:"feasible"`
+	// DurationToCreate is the duration to create the initial solution.
+	DurationToCreate float64 `json:"duration_to_create"`
```
merschformann, line 53: Is this really the performance metric? I.e., the time it took to create the initial solution? Maybe we could add the unit to the comment here (I suppose it's seconds).

nextroute/model_sanity_check.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// NewModelSanityCheck returns a new ModelSanityCheck.
+//
+// The sanity check is a function that takes a model and returns a
+// ModelSanityCheck. The sanity check is run on the model. The sanity check
+// should return the sanity check output and an error if the sanity check
+// itself fails.
+//
+//	 func someSolver(model nextroute.Model) (runSchema.Output, error) {
```
merschformann, line 143: If the sanity check does not take long it would also be advisable to add it prior to the actual run, right? Should we point that out too?

nextroute/model_sanity_check.go
```diff
@@ -0,0 +1,163 @@
...
+// ModelSanityCheck is the interface for the sanity check.
+type ModelSanityCheck interface {
+	// Check runs the sanity check. It returns the sanity check output and an
+	// error if the sanity check itself fails. Duration is the maximum time
+	// allowed for the sanity check to run. If the sanity check takes longer
+	// the remark will be "timeout". Depth is the depth of the sanity check.
+	Check(duration time.Duration, depth SanityCheckDepth) (SanityCheckOutput, error)
+}
+
+// SanityCheckOutput is the output of the sanity check.
```
merschformann, line 31: Should some of the output struct fields below here be declared with a `omitempty`? Or will they always all be present and relevant?

merschformann (approved):
Only small comments that got me thinking at first sight. I did not review the implementation yet. So, take it as user comments. :wink: 
However, nothing blocking.

PR #339, opened 2023-08-11
Improve location efficiency

dirkschumacher:
Removing the `bool` from the location struct gives us 8 bytes back (one for the bool, 7 padding).
In addition I reduced allocation overhead for unique quite a bit, made Centroid more efficient. The reduced size of the Location struct now also makes Haversine 20% faster. However this is a breaking change in theory as we would need to replace the interface with a struct

Still some further integration tests needed though.

```
name              old time/op    new time/op    delta
Distance-8          0.31ns ± 1%    0.31ns ± 0%      ~     (p=0.118 n=10+9)
Haversine-8         39.7ns ± 0%    31.3ns ± 0%   -21.34%  (p=0.000 n=9+9)
Location-8          2.03ns ± 0%    0.78ns ± 0%   -61.73%  (p=0.000 n=9+8)
Unique/n=10-8       3.62µs ± 0%    0.46µs ± 0%   -87.29%  (p=0.000 n=9+8)
Unique/n=100-8      40.9µs ± 0%     4.5µs ± 0%   -88.92%  (p=0.000 n=10+10)
Unique/n=1000-8      390µs ± 0%      43µs ± 0%   -88.93%  (p=0.000 n=10+9)
Unique/n=10000-8    3.95ms ± 0%    0.43ms ± 0%   -89.05%  (p=0.000 n=9+9)
Centroid-8          10.5µs ± 2%     1.8µs ± 0%   -82.67%  (p=0.000 n=10+10)
Filter-8             430ns ± 1%     426ns ± 0%    -0.92%  (p=0.000 n=10+9)

name              old alloc/op   new alloc/op   delta
Distance-8           0.00B          0.00B           ~     (all equal)
Haversine-8          0.00B          0.00B           ~     (all equal)
Location-8           0.00B          0.00B           ~     (all equal)
Unique/n=10-8       1.99kB ± 0%    0.50kB ± 0%   -74.93%  (p=0.000 n=10+10)
Unique/n=100-8      34.1kB ± 0%     7.2kB ± 0%   -78.87%  (p=0.000 n=10+10)
Unique/n=1000-8      309kB ± 0%      57kB ± 0%   -81.42%  (p=0.000 n=10+10)
Unique/n=10000-8    2.71MB ± 0%    0.49MB ± 0%   -82.06%  (p=0.000 n=10+10)
Centroid-8           24.0B ± 0%      0.0B       -100.00%  (p=0.000 n=10+10)
Filter-8            1.79kB ± 0%    1.79kB ± 0%      ~     (all equal)

name              old allocs/op  new allocs/op  delta
Distance-8            0.00           0.00           ~     (all equal)
Haversine-8           0.00           0.00           ~     (all equal)
Location-8            0.00           0.00           ~     (all equal)
Unique/n=10-8         42.0 ± 0%       4.0 ± 0%   -90.48%  (p=0.000 n=10+10)
Unique/n=100-8         410 ± 0%         4 ± 0%   -99.02%  (p=0.000 n=10+10)
Unique/n=1000-8      4.02k ± 0%     0.01k ± 0%   -99.83%  (p=0.000 n=10+10)
Unique/n=10000-8     40.1k ± 0%      0.0k ± 0%   -99.97%  (p=0.000 n=9+10)
Centroid-8            1.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
Filter-8              2.00 ± 0%      2.00 ± 0%      ~     (all equal)

### CODE REVIEW

**Review by davidrijsman — APPROVED** [review]

File: .golangci.yml [hunk 1]
```diff
@@ -344,9 +344,13 @@ issues:
       linters:
         - revive
       text: "exported: type name will be used as mip.MipOptions by other packages, and that stutters"
+    # We do not need to check for week random number generators in tests.
```
nilsbeck:
```suggestion
    # We do not need to check for weak random number generators in tests.
```

nilsbeck (commented):
I think this change is reasonable if it gives us couple of % performance improvement!

PR #341, opened 2023-08-16
Improve RangeMap performance

larsbeck:
# Description

By using `slices.Sort` (vs sort.Slice) and assigning via index (vs append) we can improve the performance of `RangeMap`.


davidrijsman (approved):
Obviously we allow to be dependent on experimental packages?

larsbeck:
> Obviously we allow to be dependent on experimental packages?

It moved to the standard library in Go1.21. I think this one is fine.

PR #342, opened 2023-08-16
preserves error type from table logic

chriskeyser:
# Summary

Preserves the error that is provided by the table function. In that logic we are differentiating between input errors, and system errors. We want that info to properly inform the user if its internal or their error.


merschformann (approved):
:rocket:

PR #344, opened 2023-08-17
Feature/eng 3312 add a marketplace app which conforms to the legacy v0 fleet

nilsbeck:
# Description

This PR adds a new converter function to easily convert a `cloud fleet input` into a `nextroute` input. It also adds a template which uses the converter before using nextroute to solve the problem.

## Changes

- nextroute/schema/fleet.go adds the converter and input structs
- nextroute/schema/fleet_test.go adds a test case for the converter


nextroute/schema/fleet.go
```diff
@@ -0,0 +1,234 @@
...
+	StopDuration            *int         `json:"stop_duration,omitempty"`
+	TargetTime              *time.Time   `json:"target_time,omitempty"`
+	EarlinessPenalty        *float64     `json:"earliness_penalty,omitempty"`
+	LatenessPenalty         *float64     `json:"lateness_penalty,omitempty"`
+	CompatibilityAttributes *[]string    `json:"compatibility_attributes,omitempty"`
+}
+
+// FleetToNextRoute takes a legacy cloud fleet input and converts it into
+// nextroute input format.
+func FleetToNextRoute(fleetInput FleetInput) (Input, error) {
```
sebastian-quintero:
```suggestion
func (fleetInput FleetInput) ToNextRoute() (Input, error) {
```

It makes sense to implement the function on the struct directly, rather than passing the schema as an argument.

nextroute/schema/fleet.go
```diff
@@ -0,0 +1,234 @@
+// Package schema provides the input and output schema for nextroute.
+package schema
+
+import (
+	"fmt"
+	"time"
+)
+
+// FleetInput schema.
+type FleetInput struct {
```
sebastian-quintero:
```suggestion
// FleetInput schema.
// DEPRECATION NOTICE: this part of the API is deprecated and is no longer maintained. It will be deleted soon. Please use [Input] instead.
type FleetInput struct {
```

For good measure let's make sure anyone seeing this will know it is deprecated.

nextroute/schema/fleet.go
```diff
@@ -0,0 +1,234 @@
...
+	Defaults       *FleetDefaults  `json:"defaults,omitempty"`
+	Vehicles       []FleetVehicle  `json:"vehicles,omitempty"`
+	Stops          []FleetStop     `json:"stops,omitempty"`
+	StopGroups     [][]string      `json:"stop_groups,omitempty"`
+	AlternateStops []FleetStop     `json:"alternate_stops,omitempty"`
+	DurationGroups []DurationGroup `json:"duration_groups,omitempty"`
+}
+
+// FleetDefaults holds the fleet input default data.
+type FleetDefaults struct {
```
sebastian-quintero:
```suggestion
// FleetDefaults holds the fleet input default data.
// DEPRECATION NOTICE: this part of the API is deprecated and is no longer maintained. It will be deleted soon. Please use [Defaults] instead.
type FleetDefaults struct {
```

nextroute/schema/fleet.go
```diff
@@ -0,0 +1,234 @@
...
+	Vehicles *FleetVehicleDefaults `json:"vehicles,omitempty"`
+	Stops    *FleetStopDefaults    `json:"stops,omitempty"`
+}
+
+// FleetVehicleDefaults holds the fleet input vehicle default data.
+type FleetVehicleDefaults struct {
```
sebastian-quintero: So on so forth about adding the deprecation notice. You just need to change the correct struct on each comment to point to `[VehicleDefaults]`, `[StopDefaults]`, etc...

nextroute/schema/fleet.go
```diff
@@ -0,0 +1,234 @@
...
+		if fleetInput.Defaults.Stops.HardWindow != nil {
+			input.Defaults.Stops.StartTimeWindow = *fleetInput.Defaults.Stops.HardWindow
+		}
+	}
+
+	// Use default values. nil values are not compatible between fleet and
+	// nextroute.
+	if fleetInput.Defaults != nil && fleetInput.Defaults.Vehicles != nil {
+		vehicleCompats = fleetInput.Defaults.Vehicles.CompatibilityAttributes
+		input.Defaults.Vehicles = &VehicleDefaults{
```
sebastian-quintero, line 155: If the `defaults` for `stops` is nil, the above `if` condition will not be true, and thus `input.Defaults = &Defaults{}` will never be called and this line will probably panic with nil pointer dereference. I think.

nextroute/schema/fleet.go
```diff
@@ -0,0 +1,234 @@
...
+	for i := range fleetInput.Vehicles {
+		fleetInput.Vehicles[i].CompatibilityAttributes =
+			append(fleetInput.Vehicles[i].CompatibilityAttributes, vehicleCompats...)
+	}
+
+	// Create vehicles with special logic for backlog legacy needs.
+	backlogStops := make(map[string]struct{})
+	vehicles := make([]Vehicle, len(fleetInput.Vehicles))
+	for i, v := range fleetInput.Vehicles {
+		v := v
```
sebastian-quintero, line 184: Wut?

templates/routing-legacy/main.go
```diff
@@ -0,0 +1,59 @@
...
+	err := runner.Run(context.Background())
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+type options struct {
+	Model  factory.Options                `json:"model,omitempty"`
+	Solve  nextroute.ParallelSolveOptions `json:"solve,omitempty"`
+	Format nextroute.FormatOptions        `json:"format,omitempty"`
```
sebastian-quintero, line 31: How are we going to handle the legacy options? Because they are different.

templates/routing-legacy/main.go
```diff
@@ -0,0 +1,59 @@
...
+		return runSchema.Output{}, err
+	}
+
+	solutions, err := solver.Solve(ctx, options.Solve)
+	if err != nil {
+		return runSchema.Output{}, err
+	}
+	last := solutions.Last()
+
+	output := factory.Format(ctx, options, solver, last)
```
sebastian-quintero, line 71: We need to make this also backwards compatible. So we need a new format function as well that transforms to legacy output.

sebastian-quintero (changes requested):
Looking good. I think it still needs more work.

nextroute/schema/fleet.go
nilsbeck: yep, fixed this locally. 👍

nextroute/schema/fleet.go
nilsbeck: this is a workaround gecause gosec complains because a pointer of v is passed and v is a loop variable. go does reuse the pointers for variable. this assignment gives each loop variable its own address.

templates/routing-legacy/main.go
nilsbeck: yes, because they are different. @larsbeck and I decided to leave them be.

sebastian-quintero: You can't just simply let them be, can you? Then it simply won't work for a customer transitioning that is setting `solver.limits.duration`....

nextroute/schema/fleet.go
sebastian-quintero: Why would a pointer of v be passed? I'm not seeing it I guess?

templates/routing-legacy/main.go
nilsbeck: yep, I guess we were too focussed on the other use case that Lars has which is only converting old files. So, I will add duration

nilsbeck: lies, I guess the converter will not handle it but we will set the duration in the app directly, if given.

templates/routing-legacy/main.go
merschformann: Yep, good catch.

nextroute/schema/fleet.go
nilsbeck: its a pointer of a field in v. but that comes to the same result. and pointer because nextroute input requires a pointer for some fields.

templates/routing-legacy/main.go
merschformann: Yeah, good catch. We need to handle the [legacy fleet input options](https://github.com/nextmv-io/cloud/blob/feature/eng-3392-implement-actions-from-on-call/wrangle/jobs/fleet/input.go#L138-L141).
However, we shouldn't be worried about the CLI options, since this is meant for users coming from legacy fleet (which is solely SAS).

merschformann: BTW, not sure what we are planning to do about the [header options](https://github.com/nextmv-io/cloud/blob/feature/eng-3392-implement-actions-from-on-call/core/pkg/data_plane/inputs/fleet.go#L80). Are we supporting them officially?

templates/routing-legacy/main.go
nilsbeck: legacy output will be a follow up. Ticket is here: https://linear.app/nextmv/issue/ENG-3702/add-legacy-output-to-nextroute

nextroute/schema/fleet.go
nilsbeck: sure

nextroute/schema/fleet.go
nilsbeck: sure

nextroute/schema/fleet.go
nilsbeck: yep

nextroute/schema/fleet.go
nilsbeck: followed it except for the solve options because there are so many nested structs that do not have a direct match. so its only "solve.Options" for those cases.

templates/routing-legacy/main.go
sebastian-quintero: These options are compiled into the binary. That binary is uploaded to the marketplace app and it is what the app uses to run. So that's why I think we definitely have to handle options here. Either here or inside the `solver` func.

templates/routing-legacy/main.go
sebastian-quintero: Ok.

nextroute/schema/fleet.go
sebastian-quintero: Kk.

templates/routing-legacy/main.go
nilsbeck: they are handled in the template that will be the app. to be more precise: only the run duration is handled, because diagram width etc is void and we cannot convert them as far as I know. they are not handled in the input itself because in nextroute they are not part of the input anymore put separate options. or am I missing something?

merschformann: Yes, all fine. Certainly, we can't support everything anymore. Just wondering whether someone was using the header options of legacy fleet. But maybe we did not officially support them anyway.
And for the input options we just need to make sure (even if they don't make sense anymore) that they are acceptable and the runner or something won't complain about additional fields.

nilsbeck: oh, yeah. that is a bummer. have not thought about this new feature. in general great but for this case not so cool :)

nextroute/schema/fleet.go
```diff
@@ -0,0 +1,278 @@
...
+	StopDuration            *int         `json:"stop_duration,omitempty"`
+	TargetTime              *time.Time   `json:"target_time,omitempty"`
+	EarlinessPenalty        *float64     `json:"earliness_penalty,omitempty"`
+	LatenessPenalty         *float64     `json:"lateness_penalty,omitempty"`
+	CompatibilityAttributes *[]string    `json:"compatibility_attributes,omitempty"`
+}
+
+// FleetVehicle holds the fleet input vehicle data.
+// DEPRECATION NOTICE: this part of the API is deprecated and is no longer
+// maintained. It will be deleted soon. Please use [Vehilce] instead.
```
sebastian-quintero:
```suggestion
// maintained. It will be deleted soon. Please use [Vehicle] instead.
```

nextroute/schema/fleet.go
```diff
@@ -0,0 +1,278 @@
...
+func (fleetInput FleetInput) ToNextRoute() (Input, error) {
+	input := Input{
+		Defaults: &Defaults{},
+	}
+	stopCompats := make([]string, 0)
+	vehicleCompats := make([]string, 0)
+	// Use default values and add special handling for CompatibilityAttributes
+	// and HardWindows.
+	if fleetInput.Defaults != nil && fleetInput.Defaults.Stops != nil {
+		input.Defaults = &Defaults{}
```
sebastian-quintero: This line can go away because you are doing this now:

```go
	input := Input{
 		Defaults: &Defaults{},
 	}
```

templates/routing-legacy/main.go
```diff
@@ -0,0 +1,65 @@
...
+	options options,
+) (runSchema.Output, error) {
+	nextrouteInput, err := input.ToNextRoute()
+	if err != nil {
+		panic(err)
+	}
+
+	opt := options.Solve
+	if input.Options != nil || input.Options.Solver != nil &&
+		input.Options.Solver.Limits != nil && input.Options.Solver.Limits.Duration != nil {
```
sebastian-quintero: Shouldn't the first operator be also an and?
```suggestion
	if input.Options != nil && input.Options.Solver != nil &&
		input.Options.Solver.Limits != nil && input.Options.Solver.Limits.Duration != nil {
```

templates/routing-legacy/main.go
larsbeck: We can deactivate the validation feature, by passing it as `nil`:

```
	runner := run.CLI(solver,
		run.InputValidate[run.CLIRunnerConfig, schema.Input, options, runSchema.Output](
			validate.JSON[schema.Input](nil),
		),
	)
```

nextroute/schema/fleet.go
nilsbeck: done

nextroute/schema/fleet.go
nilsbeck: done

templates/routing-legacy/main.go
nilsbeck: yes

templates/routing-legacy/main.go
nilsbeck: we talked about this code today and found that it should look like this:

```
runner := run.CLI(solver,
		run.InputValidate[run.CLIRunnerConfig, schema.FleetInput, options, runSchema.Output](
			nil,
		),
	)
```

sebastian-quintero (approved):
Curious to see this unfold.

PR #349, opened 2023-08-21
Use gosec with golangci-lint

dirkschumacher:
We can use `gosec` as a linter and don't need a dedicated GH workflow. In addition I updated the `golangci-lint` version in the workflow and pinned the `go` version to `1.19.3`


larsbeck (approved):
@dirkschumacher Nice! Could you also do that for platform (since you just did it here) or would you rather have someone else do that?

dirkschumacher:
> @dirkschumacher Nice! Could you also do that for platform (since you just did it here) or would you rather have someone else do that?

I am on it. But it'll take a bit more time to complete

PR #352, opened 2023-08-22
reverted adding a description string so error remains valid JSON

chriskeyser:
Reverts a change that puts an error string before valid error json, breaking json parsing of the error contents.


route/osrm/client.go
```diff
@@ -111,6 +111,13 @@ func (c *client) IgnoreEmpty(ignore bool) {
 	c.removeEmpty = ignore
 }
 
+// WithIgnoreEmpty configures if empty points should be ignored
```
sebastian-quintero:
```suggestion
// WithIgnoreEmpty configures if empty points should be ignored.
```
So the linter passes.

sebastian-quintero (approved):

route/osrm/client.go
```diff
@@ -111,6 +111,13 @@ func (c *client) IgnoreEmpty(ignore bool) {
 	c.removeEmpty = ignore
 }
 
+// WithIgnoreEmpty configures if empty points should be ignored.
+func WithIgnoreEmpty(ignore bool) ClientOption {
+	return func(c *client) {
+		c.removeEmpty = ignore
+	}
+}
```
merschformann, line 119: Do we need this? :thinking:

sebastian-quintero:
Outdated

PR #353, opened 2023-08-23
Feature/eng 3471 add the concept of a metaplanunit

davidrijsman:
Introduced a meta plan unit which is a composition of plan units.


sebastian-quintero (commented):
A lot of breaking changes. We are going to break anyone who has a custom constraint in Prod.

sebastian-quintero (approved):

PR #356, opened 2023-08-28
feature/req-303-workforce-scheduling-marketplace-app

nilsbeck:
# Description

Replaces a stored based shift scheduling template with a MIP based one.

# Changes
- Removes old template files
- Adds new files


templates/shift-scheduling/schema.go
```diff
@@ -0,0 +1,114 @@
...
+	DemandsCovered []RequiredWorker `json:"demands_covered"`
+	Start          time.Time        `json:"start"`
+	End            time.Time        `json:"end"`
+	Worker         Worker           `json:"worker"`
+	Duration       time.Duration    `json:"duration"`
+	AssignmentID   int              `json:"assignment_id"`
+}
+
+// DurationApart calculates the time to assignments are apart from each other.
+func (a Assignment) DurationApart(other Assignment) time.Duration {
```
dirkschumacher: That receiver should better be a pointer I think

templates/shift-scheduling/schema.go
```diff
@@ -0,0 +1,114 @@
...
+		return a.Start.Sub(other.End)
+	}
+	if a.End.Before(other.Start) {
+		return other.Start.Sub(a.End)
+	}
+	return 0
+}
+
+// ID returns the assignment id.
+func (a Assignment) ID() string {
```
dirkschumacher: That receiver should better be a pointer I think

templates/shift-scheduling/schema.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// CustomTime represents a time.Time.
+type CustomTime struct {
+	time.Time
+}
+
+// UnmarshalJSON unmarshals a CustomTime.
+func (t *CustomTime) UnmarshalJSON(b []byte) (err error) {
+	date, err := time.Parse(`"2006-01-02 15:04:05"`, string(b))
```
dirkschumacher: What is the reason here not to use RFC3339?

templates/shift-scheduling/main.go
```diff
@@ -1,370 +1,248 @@
...
-type workerID int
-
-func solver(input schedulingProblem, opts store.Options) (store.Solver, error) {
-	// We start with an empty schedule.
-	schedule := store.New()
+func solver(_ context.Context, input Input, opts Options) (out schema.Output, retErr error) {
+	// We solve a shift coverage problem using Mixed Integer Programming.
+	// We solve this by generating all possible shifts
+	// and then selecting a subset of these
+	potentialAssignments := []Assignment{}
```
dirkschumacher: Can we initialize that slice with a good capacity? 🤔

templates/shift-scheduling/main.go
```diff
@@ -1,370 +1,248 @@
...
+	potentialAssignmentsPerDriver := map[int][]Assignment{}
+	for _, driver := range input.Workers {
```
dirkschumacher: `driver` => `worker`

templates/shift-scheduling/main.go
```diff
@@ -1,370 +1,248 @@
...
+		potentialAssignmentsPerDriver[driver.ID] = []Assignment{}
+		for _, availability := range driver.Availability {
+			for start := availability.Start.Time; start.Before(availability.End.Time); start = start.Add(30 * time.Minute) {
```
dirkschumacher: I think these two for loops require at least a comment that they produce these 30 minute chunks. Or maybe better to extract the assignment generation code into its own function so that `solve` is easier to parse

templates/shift-scheduling/main.go
```diff
@@ -1,370 +1,248 @@
...
-
-func solver(input schedulingProblem, opts store.Options) (store.Solver, error) {
-	// We start with an empty schedule.
-	schedule := store.New()
+func solver(_ context.Context, input Input, opts Options) (out schema.Output, retErr error) {
+	// We solve a shift coverage problem using Mixed Integer Programming.
+	// We solve this by generating all possible shifts
+	// and then selecting a subset of these
+	potentialAssignments := []Assignment{}
+	potentialAssignmentsPerDriver := map[int][]Assignment{}
```
dirkschumacher: We store assignments here twice. The map could be `map[int][]*Assignment{}`. Or we drop `potentialAssignments` and use `potentialAssignmentsPerDriver` for everything

templates/shift-scheduling/main.go
```diff
@@ -1,370 +1,248 @@
...
 
-	// We define some helper variables that are useful later.
-	nShifts := input.Days * 3
-	workers := model.NewDomain(model.NewRange(1, input.Workers))
-	typeMap := [3]shiftType{"morning", "day", "night"}
-	prefShiftType := map[int]shiftType{}
-	for _, v := range input.Preferences.ShiftType {
-		prefShiftType[int(v.Worker)] = v.Type
+	// initialize demand ids
+	for i, demand := range input.RequiredWorkers {
```
dirkschumacher: This overrides the input? Maybe we should assign worker IDs outside of the input struct.

templates/shift-scheduling/main.go
```diff
@@ -1,370 +1,248 @@
...
-		for _, d := range v.Days {
-			morning := (d - 1) * 3
-			day := morning + 1
-			night := day + 1
-			changes = append(changes,
-				shifts.Remove(morning, []int{int(v.Worker)}),
-				shifts.Remove(day, []int{int(v.Worker)}),
-				shifts.Remove(night, []int{int(v.Worker)}),
-			)
+	demandCovering := map[int][]Assignment{}
```
dirkschumacher: I think we could store pointers here to not duplicate assignments again

templates/shift-scheduling/main.go
```diff
@@ -1,370 +1,248 @@
...
+	for _, demand := range input.RequiredWorkers {
```
dirkschumacher: These for loops could also be extracted into a helper function to make `solve` easier to read.

e.g. 
```go
func solve(...) {
  availableShifts := generateAvailableShifts(...)
  demand := generateDemand(...)
  model := buildMIPModel(input, availableShifts, demand)
  ...
}

```

templates/shift-scheduling/main.go
```diff
@@ -1,370 +1,248 @@
...
-	// Now we define our constraint containing a propagator. This will remove
-	// workers from domains based upon the workers in other domains.
-	constraint := newBreakConstraint(shifts, 2)
-
-	// We invoke the propagator to propagate the initial values.
-	schedule.Propagate(constraint.propagate)
-
-	if constraint.isProvenInfeasible(schedule) {
-		return nil, fmt.Errorf("input can not be solved")
+	m := mip.NewModel()
```
dirkschumacher, line 97: Like mentioned above let's extract the mip model building into a private helper function

templates/shift-scheduling/main.go
```diff
@@ -1,370 +1,248 @@
...
+		func(demand ...RequiredWorker) mip.Float {
+			return m.NewFloat(0, float64(demand[0].Count))
+		}, input.RequiredWorkers)
+
+	overSupplySlack := model.NewMultiMap(
+		func(demand ...RequiredWorker) mip.Float {
+			return m.NewFloat(0, math.MaxFloat64)
+		}, input.RequiredWorkers)
+
+	for _, demand := range input.RequiredWorkers {
```
dirkschumacher, line 115: Why don't we loop directly over `demandCovering`?

templates/shift-scheduling/main.go
```diff
@@ -1,370 +1,248 @@
...
+		demandCover := demandCovering[demand.RequiredWorkerID]
+		// We need to cover all demands
+		coverConstraint := m.NewConstraint(mip.Equal, float64(demand.Count))
+		coverConstraint.NewTerm(1.0, underSupplySlack.Get(demand))
+		coverConstraint.NewTerm(-1.0, overSupplySlack.Get(demand))
+		coverPerDriver := map[int]mip.Constraint{}
```
dirkschumacher: driver? Should be called worker I guess

templates/shift-scheduling/main.go
```diff
@@ -1,370 +1,248 @@
...
+		for _, assignment := range demandCover {
+			if constraint, ok := coverPerDriver[assignment.Worker.ID]; !ok {
+				// When two potential assignments of a driver overlap, only one
+				// can be assigned
+				coverPerDriver[assignment.Worker.ID] = m.NewConstraint(mip.LessThanOrEqual, 1.0)
+				coverPerDriver[assignment.Worker.ID].NewTerm(1.0, x.Get(assignment))
+			} else {
+				constraint.NewTerm(1.0, x.Get(assignment))
```
dirkschumacher: Does that actually work? I guess 🤔
Maybe better to do

```go
constraint, ok := coverPerDriver[assignment.Worker.ID]
if !ok {
    constraint = m.NewConstraint(mip.LessThanOrEqual, 1.0)
    coverPerDriver[assignment.Worker.ID] = constraint
}
constraint.NewTerm(1.0, x.Get(assignment))


```

templates/shift-scheduling/main.go
```diff
@@ -1,370 +1,248 @@
...
-		i, ok := shifts.First(s)
-		workers := shifts.Domain(s, i).Slice()
-
-		return store.Lazy(
-			func() bool {
-				for j := 0; j < nShifts; j++ {
-					if shifts.Domain(s, j).Empty() {
-						return false
+	// Two shift of a driver have to be at least 8 hours apart
+	for _, driver := range input.Workers {
```
dirkschumacher: driver?

templates/shift-scheduling/main.go
```diff
@@ -1,370 +1,248 @@
...
+		for i, a1 := range potentialAssignmentsPerDriver[driver.ID] {
+			// A driver can only work 10 hours per day
```
dirkschumacher: We don't have the concept of drivers in the schema. We should call them workers

dirkschumacher: That comment is wrong since we let the user parameterize the problem

templates/shift-scheduling/main.go
```diff
@@ -1,370 +1,248 @@
...
+			lessThan10hPerDay := m.NewConstraint(mip.LessThanOrEqual, opts.MaxHoursPerDay.Hours())
```
dirkschumacher: The variable names have fixed lengths in them (e.g. 10 hours or 8 hours or 40h) but we these values are user parameters.

templates/shift-scheduling/main.go
```diff
@@ -1,370 +1,248 @@
...
-		if wcLower < 3 {
-			wcLower = 3
-		}
+	solver, err := mip.NewSolver("highs", m)
+	if err != nil {
+		return schema.Output{}, err
+	}
+	options := mip.NewSolveOptions()
+	options.SetVerbosity(mip.Off)
+	err = options.SetMIPGapRelative(0.999)
```
dirkschumacher: That gap could be a `const` variable somewhere at the top of the file

templates/shift-scheduling/main.go
```diff
@@ -1,370 +1,248 @@
...
-				if i+j < t.shifts.Len(s) && t.checkFeasible(s, w1, i+j) {
-					return true
+		nextShiftSolution.Value = solution.ObjectiveValue()
+		val := statistics.Float64(solution.ObjectiveValue())
+		result.Value = &val
+
+		usedWorkers := make(map[int]struct{})
+
+		for _, assignment := range assignments {
+			if solution.Value(x.Get(assignment)) >= 0.5 {
```
dirkschumacher: `>= 0.9` is probably a bit close to what we want to do (protect against small numerical problems)

dirkschumacher (changes requested):
Great. Some first comments

templates/shift-scheduling/schema.go
nilsbeck: here it could be a receiver but since the other one cannot be a receiver...keep them both not a receiver? I think that sounds better?

templates/shift-scheduling/schema.go
nilsbeck: this cannot be a receiver because it is used in a MultiMap and needs to comply with the mode.Identifier interface. Turning it into a receiver breaks that.

templates/shift-scheduling/schema.go
nilsbeck: I guess there is no reason, changed it.

templates/shift-scheduling/main.go
nilsbeck: I don't think so because the idea here is to reduce the total assignements options by looking at the availabilities and so on.

templates/shift-scheduling/main.go
nilsbeck: yes

templates/shift-scheduling/main.go
nilsbeck: replaced all notions of driver with worker

nilsbeck: fixed with variables :)

templates/shift-scheduling/main.go
nilsbeck: changed for variables

templates/shift-scheduling/main.go
nilsbeck: sure

templates/shift-scheduling/main.go
nilsbeck: changed this as well.

dirkschumacher:
@nilsbeck there are some unresolved comments from mine. Have you seen them? The GH UI hid them 🤔

templates/shift-scheduling/schema.go
dirkschumacher: Seems still be there 🤔

dirkschumacher (changes requested):
@nilsbeck there are some unresolved comments from mine. Have you seen them? The GH UI hid them 🤔

templates/shift-scheduling/main.go
nilsbeck: I think it's easier to work with the enties IDs and not ints directly?

templates/shift-scheduling/main.go
nilsbeck: results in the same. fine with me

templates/shift-scheduling/main.go
nilsbeck: this does not seem to work. would need to investigate a bit more where this actually breaks

nilsbeck:
> @nilsbeck there are some unresolved comments from mine. Have you seen them? The GH UI hid them 🤔

had not seen them :( but I worked through them now.

templates/shift-scheduling/input.json
```diff
@@ -1,42 +1,88 @@
...
-      "days": [1]
-    },
-    {
-      "worker": 4,
-      "days": [4, 5]
-    },
-    {
-      "worker": 5,
-      "days": [4, 5]
+  "workers": [
```
sebastian-quintero, line 2: This seems like a weird input to me. You have 5 workers, yet you require a minimum of 16 workers to cover the shifts, given the `required_workers` `count`.

templates/shift-scheduling/schema.go
```diff
@@ -0,0 +1,114 @@
+package main
+
+import (
+	"strconv"
+	"time"
+)
+
+// Output holds the output data of the solution.
+type Output struct {
```
sebastian-quintero: There is no need for any of the structs to be exported.

templates/shift-scheduling/schema.go
sebastian-quintero: Same comment.

templates/shift-scheduling/schema.go
```diff
@@ -0,0 +1,114 @@
...
+	t.Time = date
+	return
+}
+
+// MarshalJSON marshals a CustomTime.
+func (t *CustomTime) MarshalJSON() ([]byte, error) {
+	return []byte(t.Format(time.RFC3339)), nil
+}
+
+// OutputAssignment holds an assignment for a driver.
```
sebastian-quintero:
```suggestion
// OutputAssignment holds an assignment for a worker.
```

templates/shift-scheduling/output.json
```diff
@@ -0,0 +1,35 @@
+{
```
sebastian-quintero: The output does not hold the same structure as the other `mip` templates. It would be good to keep the same structure for consistency.

Check out how, for example, the knapsack example looks like:

```json
{
  "options": {
    "limits": {
      "duration": 10000000000
    }
  },
  "solutions": [
    {
      "items": [
        {
          "item_id": "cat",
          "value": 100,
          "weight": 20
        },
        {
          "item_id": "water",
          "value": 40,
          "weight": 2
        },
        {
          "item_id": "phone",
          "value": 6,
          "weight": 1
        },
        {
          "item_id": "book",
          "value": 63,
          "weight": 10
        },
        {
          "item_id": "rx",
          "value": 81,
          "weight": 1
        },
        {
          "item_id": "coat",
          "value": 44,
          "weight": 9
        },
        {
          "item_id": "keys",
          "value": 92,
          "weight": 1
        },
        {
          "item_id": "nuts",
          "value": 18,
          "weight": 4
        }
      ]
    }
  ],
  "statistics": {
    "result": {
      "custom": {
        "constraints": 1,
        "provider": "highs",
        "status": "optimal",
        "variables": 11
      },
      "duration": 0.000540542,
      "value": 444
    },
    "run": {
      "duration": 0.000540542
    },
    "schema": "v1"
  },
  "version": {
    "sdk": "VERSION"
  }
}
```

Also, not sure why we need this file here?

templates/shift-scheduling/output.json
```diff
@@ -0,0 +1,35 @@
...
+  "version": { "sdk": "v0.28.0" },
+  "solutions": [
+    {
+      "status": "optimal",
```
sebastian-quintero: I don't see how this solution satisfies the constraints. The shifts are not being covered with the required number of workers...

templates/shift-scheduling/main.go
```diff
@@ -1,370 +1,268 @@
...
+	}
 
-	// We invoke the propagator to propagate the initial values.
-	schedule.Propagate(constraint.propagate)
+	return output, nil
+}
 
-	if constraint.isProvenInfeasible(schedule) {
-		return nil, fmt.Errorf("input can not be solved")
+func format(
```
sebastian-quintero, line 62: I now understand why the output looks different. Can you please look at the other MIP templates and copy what is being used as the `format` function? There is some functionality in `sdk` that already takes care of setting the value, optimality, etc... It would be good to be consistent.

Take this example from the `mip-knapsack` template:

```go
	// Format the solution into the desired output format and add custom
	// statistics.
	output := mip.Format(options, format(input, solution, itemVariables), solution)
	output.Statistics.Result.Custom = mip.DefaultCustomResultStatistics(m, solution)

	return output, nil
}

// format the solution from the solver into the desired output format.
func format(input input, solverSolution mip.Solution, itemVariables map[string]mip.Bool) solution {
	if !solverSolution.IsOptimal() && !solverSolution.IsSubOptimal() {
		return solution{}
	}

	items := make([]item, 0)
	for _, item := range input.Items {
		if solverSolution.Value(itemVariables[item.ItemID]) < 1 {
			continue
		}
		items = append(items, item)
	}

	return solution{
		Items: items,
	}
}
```

You can see that `format` _just_ takes care of translating the `mip.Solution` into the specific `solution` of the template. The rest of the logic is taken care of by these 2 lines:

```go
	output := mip.Format(options, format(input, solution, itemVariables), solution)
	output.Statistics.Result.Custom = mip.DefaultCustomResultStatistics(m, solution)
```

The `mip.Format` function is used...

templates/shift-scheduling/output.json
sebastian-quintero: I see, so the demand constraint has slack variables. It would be helpful to print out the value of these slack variables for each shift that is missing. Basically, would be useful to know, in the same manner as we have the unplanned stops, how many workers are missing for shifts that are undersupplied (or viceversa).

sebastian-quintero (dismissed):
There are some changes around the output specifically that I believe should be there...

templates/shift-scheduling/schema.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+// output holds the output data of the solution.
+type output struct {
+	AssignedShifts        []outputAssignment `json:"assigned_shifts"`
+	NumberAssignedWorkers int                `json:"number_assigned_workers"`
+}
+
+// options holds custom configuration data.
+type options struct {
```
dirkschumacher, line 14: Would be great to think about a possible nested structure here to have nicer looking CMD args.
E.g.
```go
penalty {
  oversupply
  undersupply
}
```
results in `-penalty.oversupply=1000` which looks nicer than `-oversupplypenalty=1000` IMO

templates/shift-scheduling/schema.go
```diff
@@ -0,0 +1,109 @@
...
+	OverSupplyPenalty  float64       `json:"over_supply_penalty" default:"1000" usage:"penalty for over-supplying a demand"`
+	UnderSupplyPenalty float64       `json:"under_supply_penalty" default:"500" usage:"penalty for over-supplying a demand"`
+	MaxHoursPerDay     time.Duration `json:"max_hours_per_day" default:"10h" usage:"maximum number of hours per day"`
+	MaxHoursPerWeek    time.Duration `json:"max_hours_per_week" default:"40h" usage:"maximum number of hours per week"`
+	MinHoursPerShift   time.Duration `json:"min_hours_per_shift" default:"2h" usage:"minimum number of hours per shift"`
+	MaxHoursPerShift   time.Duration `json:"max_hours_per_shift" default:"8h" usage:"maximum number of hours per shift"`
+	HoursBetweenShifts time.Duration `json:"hours_between_shifts" default:"8h" usage:"minimum number of hours between shifts"`
+	Limits             mip.Limits    `json:"limits,omitempty"`
```
dirkschumacher: Can you add a comment how the limit is supposed to be used

templates/shift-scheduling/schema.go
```diff
@@ -0,0 +1,109 @@
...
+	Count            int        `json:"count"`
+}
+
+// ID returned the RequiredWorker ID.
+func (r requiredWorker) ID() string {
+	return strconv.Itoa(r.RequiredWorkerID)
+}
+
+// customTime represents a time.Time.
+type customTime struct {
```
dirkschumacher: Why do we need a custom time? And if we need it, why not do `type customTime time.Time`?

templates/shift-scheduling/main.go
```diff
@@ -1,370 +1,225 @@
...
+	nextShiftSolution.NumberAssignedWorkers = len(usedWorkers)
+
+	return nextShiftSolution
+}
+
+func newMIPModel(
+	input input,
+	potentialAssignments []assignment,
+	potentialAssignmentsPerWorker map[int][]*assignment,
+	demandCovering map[int][]assignment,
```
dirkschumacher: Why is `potentialAssignmentsPerWorker` a `map[int][]*assignment` and `demandCovering` a `map[int][]assignment`. Can we harmonize that?

templates/shift-scheduling/schema.go
```diff
@@ -0,0 +1,109 @@
...
+	DemandsCovered []requiredWorker `json:"demands_covered"`
+	Start          time.Time        `json:"start"`
+	End            time.Time        `json:"end"`
+	Worker         worker           `json:"worker"`
+	Duration       time.Duration    `json:"duration"`
+	AssignmentID   int              `json:"assignment_id"`
+}
+
+// DurationApart calculates the time to assignments are apart from each other.
+func (a *assignment) DurationApart(other assignment) time.Duration {
```
dirkschumacher: If it uses a pointer receiver then `other` should also be a pointer IMO.

templates/shift-scheduling/main.go
```diff
@@ -1,370 +1,225 @@
...
 	"time"
 
+	"github.com/nextmv-io/sdk/mip"
 	"github.com/nextmv-io/sdk/model"
 	"github.com/nextmv-io/sdk/run"
-	"github.com/nextmv-io/sdk/store"
+	"github.com/nextmv-io/sdk/run/schema"
 )
 
+const gap = 0.999
```
dirkschumacher: Can we rename that to `mipGap` and

templates/shift-scheduling/schema.go
```diff
@@ -0,0 +1,109 @@
...
+	return strconv.Itoa(r.RequiredWorkerID)
+}
+
+// customTime represents a time.Time.
+type customTime struct {
+	time.Time
+}
+
+// UnmarshalJSON unmarshals a CustomTime.
+func (t *customTime) UnmarshalJSON(b []byte) (err error) {
```
dirkschumacher: Should be a value receiver because `time.Time.UnmarshalJSON` also uses a value receiver

templates/shift-scheduling/schema.go
```diff
@@ -0,0 +1,109 @@
...
+	err = t.Time.UnmarshalJSON(b)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// MarshalJSON marshals a CustomTime.
+func (t *customTime) MarshalJSON() ([]byte, error) {
```
dirkschumacher: Value receiver, same reasoning as above

dirkschumacher (commented):
Almost there. Some minor comments. Some comments by @sebastian-quintero are undressed I believe.

templates/shift-scheduling/schema.go
```diff
@@ -0,0 +1,93 @@
...
+
+// options holds custom configuration data.
+type options struct {
+	Penalty            penalty       `json:"penalty" usage:"set penalties for over and under supply of workers"`
+	MaxHoursPerDay     time.Duration `json:"max_hours_per_day" default:"10h" usage:"maximum number of hours per day"`
+	MaxHoursPerWeek    time.Duration `json:"max_hours_per_week" default:"40h" usage:"maximum number of hours per week"`
+	MinHoursPerShift   time.Duration `json:"min_hours_per_shift" default:"2h" usage:"minimum number of hours per shift"`
+	MaxHoursPerShift   time.Duration `json:"max_hours_per_shift" default:"8h" usage:"maximum number of hours per shift"`
+	HoursBetweenShifts time.Duration `json:"hours_between_shifts" default:"8h" usage:"minimum number of hours between shifts"`
+	Limits             mip.Limits    `json:"limits" usage:"holds a field to set the maximum duration of the solver"`
```
dirkschumacher: let's call that `solve.duration` (nested) like in nextroute

templates/shift-scheduling/schema.go
dirkschumacher: @nilsbeck do you think nesting the other options makes sense as well? I don't have a good suggestion though 🤔

```
Limits {
   Shift {
       MinHours int
       MaxHours int
   }
   Week {
      MaxHours int
   }
   Day {
      MaxHours int
   }
}
```

In addition: why is the type `time.Duration` and not `int`?

dirkschumacher (changes requested):
Small comment/question around options and one little request to rename the `Limits` to `solve.duration` as in nextroute (if makes sense)

templates/shift-scheduling/schema.go
nilsbeck: this is because all mip model use that Limits struct. so they are all aligned (and they were with router as well, I believe). this was also a wish sebas had.

nilsbeck:
According to bug bash one feedback was that the shift limits were not that bad and didn't bother too much.
I think there is no specific reason other than that you can easily read the input and just have different time units to express your stuff. we use time.Duration for other times related things as well? but here it is hours already in the field name. which makes sense because we are talking shifts....express a shift in seconds or minutes is odd.

dirkschumacher (approved):

PR #357, opened 2023-08-28
Enable the use of the checker.

davidrijsman:
Allow the user to enable a sanity check on the model


nextroute/model_sanity_check.go
```diff
@@ -161,3 +171,13 @@ func NewModelSanityCheck(model Model) ModelSanityCheck {
 	connect.Connect(con, &newModelSanityCheck)
 	return newModelSanityCheck(model)
 }
+
+// SanityCheckReport is the sanity check of a model.
+func SanityCheckReport(
+	ctx context.Context,
+	model Model,
+	duration time.Duration,
+	sanityCheckDepth SanityCheckDepth) (schema.Output, error) {
```
sebastian-quintero:
```suggestion
	sanityCheckDepth SanityCheckDepth,
) (schema.Output, error) {
```

Since all the args are in a new line, I would prefer this. Very subjective though.

templates/nextroute/main.go
```diff
@@ -36,6 +37,15 @@ func solver(
 		return runSchema.Output{}, err
 	}
 
+	if options.SanityCheck.Enable {
```
sebastian-quintero: Why do we need to add this to the template? Can't we check if the option is enabled on the factory side in platform and if it is then add the sanity check to the output?

davidrijsman: We do not **add** the  sanity check but enable it excluding solving. The sanity check can take a lot of time (away from solving).

nextroute/model_sanity_check.go
```diff
@@ -1,11 +1,21 @@
...
 	"time"
 
 	"github.com/nextmv-io/sdk/connect"
+	"github.com/nextmv-io/sdk/run/schema"
 )
 
+// SanityCheckOptions are the options for the sanity check.
+type SanityCheckOptions struct {
+	Enable bool `json:"enable" usage:"enable the sanity check" default:"false"`
+	// Duration is the maximum duration allowed for the sanity check to run.
```
dirkschumacher: I think the comment can go away, as we already have a `usage` field

dirkschumacher (approved):

nextroute/schema/output.go
```diff
@@ -2,13 +2,16 @@ package schema
...
 
 // SolutionOutput represents a solutions as JSON.
 type SolutionOutput struct {
-	Unplanned []StopOutput    `json:"unplanned"`
-	Vehicles  []VehicleOutput `json:"vehicles"`
-	Objective ObjectiveOutput `json:"objective"`
+	Unplanned []StopOutput               `json:"unplanned"`
+	Vehicles  []VehicleOutput            `json:"vehicles"`
+	Objective ObjectiveOutput            `json:"objective"`
+	NextCheck *nextroute.NextCheckOutput `json:"next_check,omitempty"`
```
sebastian-quintero: This is a good as a place as any for this comment so here I go.

I think the API can be improved a bit. All exported symbols are named `NextCheck...`. This hints to the fact that it belongs in its own package. I would create a package under `nextroute` and define all the symbols there, minus the `NextCheck...` prefix of course.

Instead of `nextroute.NextCheckOutput` it then becomes `package.Output`, which is more aligned with Go design guides of placing stuff that belongs together in a package.

Regarding the package name, I'm not a huge fan of proprietary names. In this case, it is clear that the checker is for `nextroute`, so I don't see the necessity for adding yet another `next` word to it.

I would create the `nextroute/check` package, and thus the usage is:

- `check.Model`
- `check.Solution`
- `check.Options`
- `check.Output`

Notice in the suggestion that the type reads more smoothly.

```suggestion
	Check *check.Output `json:"check,omitempty"`
```

nextroute/schema/output.go
```diff
@@ -2,13 +2,16 @@ package schema
 
 import (
 	"time"
+
+	"github.com/nextmv-io/sdk/nextroute"
```
sebastian-quintero: Linked to my comment about placing the check stuff in its own package, notice here that `schema` is importing `nextroute`. I consider `schema` a sub-package of `nextroute` so I find it strange that `schema` has to import its parent package. If anything, it should be the other way around, where necessary.

nextroute/factory/format.go
```diff
@@ -23,6 +23,19 @@ func Format(
 	return format(ctx, options, progressioner, solutions...)
 }
 
+// FormatExt formats a solution in a basic format using the [schema.Output] to
+// format a solution.
+func FormatExt(
```
sebastian-quintero: What does `Ext` mean? Do we need to keep the other `Format` function as well. Can't we just adapt the existing `Format` function?

nextroute/model_next_check.go
```diff
@@ -0,0 +1,195 @@
...
+) (NextCheckOutput, error) {
+	connect.Connect(con, &nextCheck)
+	return nextCheckSolution(ctx, solution, duration, verbosity)
+}
+
+// NextCheckOptions are the options for a check.
+type NextCheckOptions struct {
+	// Duration is the maximum duration allowed for the sanity check to run.
+	Duration  time.Duration `json:"duration" usage:"maximum duration of the sanity check" default:"30s"`
+	Verbosity int           `json:"verbosity" usage:"verbosity of the sanity check, current available options are 0 (off),1 (low), 2 (medium), 3 (high)" default:"0"`
```
sebastian-quintero: Verbosity in `mip` was changed to a `string`. I would prefer for that to be the case here as well for consistency.

sebastian-quintero (changes requested):
Some API discussion that I believe is relevant to make my comment blocking. Thanks!

templates/nextroute/main.go
```diff
@@ -21,9 +21,10 @@ func main() {
 }
 
 type options struct {
-	Model  factory.Options                `json:"model,omitempty"`
-	Solve  nextroute.ParallelSolveOptions `json:"solve,omitempty"`
-	Format nextroute.FormatOptions        `json:"format,omitempty"`
+	Model     factory.Options                `json:"model,omitempty"`
+	Solve     nextroute.ParallelSolveOptions `json:"solve,omitempty"`
+	Format    nextroute.FormatOptions        `json:"format,omitempty"`
+	NextCheck nextroute.NextCheckOptions     `json:"next_check,omitempty"`
```
sebastian-quintero:
```suggestion
	Check check.Options     `json:"check,omitempty"`
```

If we organize the checker things into its own package, you avoid the next-next stuttering here. _Especially_ for options I find `check.Options` to be the options for the check.... Doesn't get simpler than that.

nextroute/model_sanity_check.go
davidrijsman: It is gone

nextroute/schema/output.go
davidrijsman: Makes perfect sense, have created a package and moved stuff, changed verbosity option to string

nextroute/check/check.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"context"
+
+	"github.com/nextmv-io/sdk/connect"
+	"github.com/nextmv-io/sdk/nextroute"
+)
+
+// ModelCheck is the check of a model returning a [Output].
+func ModelCheck(
```
sebastian-quintero, line 11:
```suggestion
func Model(
```

I would consider renaming this. When you are using it elsewhere, it will be `check.ModelCheck` which has a stutter. Instead, `check.Model` is just as clear.

nextroute/check/check.go
```diff
@@ -0,0 +1,28 @@
...
+	ctx context.Context,
+	model nextroute.Model,
+	options Options,
+) (Output, error) {
+	connect.Connect(con, &modelCheck)
+	return modelCheck(ctx, model, options)
+}
+
+// SolutionCheck is the check of a solution returning a [Output].
+func SolutionCheck(
```
sebastian-quintero, line 21:
```suggestion
func Solution(
```

Same reasoning here

nextroute/check/options.go
```diff
@@ -0,0 +1,9 @@
+package check
+
+import "time"
+
+// Options are the options for a check.
+type Options struct {
+	Duration  time.Duration `json:"duration" usage:"maximum duration of the check" default:"30s"`
+	Verbosity string        `json:"verbosity" usage:"verbosity of the check, current available options are 'off' , 'low', 'medium', 'high', 'very_high'" default:"off"`
```
sebastian-quintero:
```suggestion
	Verbosity string        `json:"verbosity" usage:"{off, low, high, very_high} verbosity of the check" default:"off"`
```

This is how we display set member with solutions all/last and verbosity in `mip` as well.

nextroute/check/schema.go
```diff
@@ -0,0 +1,202 @@
...
+	High
+	// VeryHigh reports for each plan unit which vehicle have moves.
+	VeryHigh
+)
+
+// ToVerbosity converts a string to a verbosity. The string can be
+// anything that starts with "o", "l", "m", "h" or "v" case-insensitive.
+// If the string does not start with one of these characters the
+// verbosity is off.
+func ToVerbosity(s string) Verbosity {
```
sebastian-quintero, line 26: Aren't we being to flexible here? Wouldn't it be better to now swallow user errors (typos) and instead make sure the string is actually correct?

sebastian-quintero (approved):
Thank you for marking the API change, I think it looks much cleaner now!

nextroute/check/check.go
davidrijsman: Model is not really a verb... and if it would be it does not communicate what the method does. API should communicate what it does (not the package name). Unfortunately Go has no function/method overloading so we can not use the API Check which would have been appropriate. Also we can not use CheckModel as Linter does not allow the method to start with the package name....

PR #360, opened 2023-08-30
Adds app manifest to ortools template

merschformann:
# Description

Adds missing app manifest to ortools template.


templates/ortools/app.yaml
```diff
@@ -0,0 +1,4 @@
+type: python
+runtime: ghcr.io/nextmv-io/runtime/ortools:latest
+files:
+  - main.py
```
sebastian-quintero, line 4: Don't we need the requirements.txt too?

sebastian-quintero (approved):

templates/ortools/app.yaml
merschformann: We can spare it. We are not installing (we can't even) anything on platform side. So, it's useless there. It's good to have it for the user though.
That reminds me, I also quickly aligned the ortools version. We need to specify the one we use in the backend.

PR #361, opened 2023-09-01
feature/eng-3786-add-state-field-to-rounting-legacy

nilsbeck:
# Description

Adds missing field `state`. Also makes use of new `All()` function.

## Changes

* `output.go`
* `main.go`


templates/routing-legacy/main.go
```diff
@@ -74,15 +74,13 @@ func solver(
 		return FleetOutput{}, err
 	}
 
-	runSolutions := run.Last
```
merschformann: I am not completely following why we change the default output type of the routing legacy app. The issue doesn't have context either. Can you catch me up? :thinking:

merschformann (approved):
Just a question about the change in default, but maybe that was the point and I just don't have the info. :relaxed:

templates/routing-legacy/main.go
nilsbeck: I think, I thought for a moment, that we returned all solutions by default in cloud. but that is incorrect. so I reverted this change.

merschformann: Oh yeah, makes sense. We did return all solutions for the frontend to visualize them. But in normal use, we did not. :relaxed:

PR #362, opened 2023-09-06
ENG-3806 Add app.yaml file to templates and move old templates to legacy

sebastian-quintero:
# Description

Only keep these templates in the `templates` location:

- `nextroute`
- `mip` 👉🏻 the same as `mip-knapsack` but just renamed to `mip`.
- `ortools`
- `shift-scheduling`

These templates have the `app.yaml` manifest file added to them. The `legacy/routing-legacy` template also receives the manifest because this template is being maintained.

The rest of the templates are moved to `templates/legacy`.


templates/legacy/templates.go
```diff
@@ -0,0 +1,314 @@
+// Package legacy contains variables holding embded legacy template files.
```
merschformann:
```suggestion
// Package legacy contains variables holding embedded legacy template files.
```
Really irrelevant, but couldn't resist. :sweat_smile:

templates/mip/app.yaml
```diff
@@ -0,0 +1,5 @@
+# This manifest holds the information the app needs to run on the Nextmv Cloud.
+type: sdk
+features:
+  - sdk
+  - mip-highs # Replace with mip-xpress to use the mip.Xpress solver provider.
```
merschformann: Should we already advertise this here? We would need to update this too, if we introduce further solvers. :thinking:

templates/mip/main.go
```diff
@@ -0,0 +1,137 @@
...
+		// Sets the value of the item in the objective function.
+		m.Objective().NewTerm(item.Value, itemVariables[item.ItemID])
+
+		// Sets the weight of the item in the constraint.
+		capacityConstraint.NewTerm(item.Weight, itemVariables[item.ItemID])
+	}
+
+	// Create a solver using a provider. Please see the documentation on
+	// [mip.SolverProvider] for more information on the available providers.
+	solver, err := mip.NewSolver(mip.Highs, m)
```
merschformann, line 85: If we keep the comment from the yaml referring to xpress, we should potentially point out that the user also needs to modify this line according to the solver they are choosing. :shrug:

templates/ortools/app.yaml
```diff
@@ -1,4 +1,6 @@
+# This manifest holds the information the app needs to run on the Nextmv Cloud.
 type: python
 runtime: ghcr.io/nextmv-io/runtime/ortools:latest
+# List all files/directories that should be included in the app.
```
merschformann:
```suggestion
# List all files/directories that should be included in the app. Globbing (e.g.: configs/*.json) is supported.
```
We could remark it here, but it probably fits the docs more and is verbose here.

merschformann (approved):
Great cleanup! Just some minor comments. Thank you for the work! :hugs:

templates/legacy/templates.go
sebastian-quintero: Listo.

templates/mip/app.yaml
sebastian-quintero: I'll add a more generic comment.

templates/mip/main.go
sebastian-quintero: I'll leave this as is for it is fairly generic to support more solvers. I updated the comment in the yaml.

templates/ortools/app.yaml
sebastian-quintero: Listo.

cowanwalls (approved):

PR #365, opened 2023-09-12
ENG-3811 Harmonize the options for MIP problems

sebastian-quintero:
# Description

- Reworked how options are passed to `mip` templates. Deleted everything related to the `SolveOptions` interface.
- The main motivation behind this change is to align `mip` template (and the engine) with `nextroute`. Also, there was a lot of unused code because "custom" options were not connected to providers.
- Options are now set via the CLI, the same way as in `nextroute`.
- Solver-specific options are now connected to the providers. 
- All the `mip` templates were updated to reflect the new usage of options (which is simpler). The templates were also reorganized a bit for more structure.
- Last, but not least, changed `-limits.duration` to `-solve.duration` (the main goal of this ticket hahaha).

## Functionality

```bash
nextmv sdk run . -- `
  -runner.input.path input.json \
  -runner.output.path output.json \
  -solve.duration 10s
```

Because there are some options that are parsed by the runner, one can now set the verbosity and MIP gaps directly with the CLI:

```bash
nextmv sdk run . -- `
  -runner.input.path input.json \
  -runner.output.path output.json \
  -solve.duration 10s \
  -solve.verbosity high \
  -solve.mip.gap.absolute 100 \
  -solve.mip.gap.relative 0.1
```

## HiGHS options

Given that the `mip.Highs`provider is now fully connected with `control` options, one can now use any [HiGHS option](https://ergo-code.github.io/HiGHS/dev/options/definitions/) directly in the CLI. You just have to know the appropriate name and type. You can pass in as many options as desired of the form `name1=value1,name2=value2`.

```bash
nextmv sdk run . -- `
  -runner.input.path input.json \
  -runner.output.path output.json \
  -solve.duration 10s \
  -solve.verbosity high \
  -solve.mip.gap.absolute 100 \
  -solve.mip.gap.relative 0.1 \
  -solve.control.float "infinite_cost=10000,infinite_bound=12345" \
  -solve.contro.int "threads=3"
```

Setting options using `-solve.control...` overrides the "native" options like `-solve.duration` for the same option. That means that if you use `-solve.duration 10s` and `-solve.control.float "time_limit=4"` the solver will run for 4 seconds.

## Xpress options

The `mip.Xpress` provider is also now fully connected with `control` options, the same way as HiGHS. One can use any [Xpress contol option](https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/HTML/chapter7.html). The behavior is the same as in HiGHS, you just need to know the names and types:

```bash
nextmv sdk run . -- `
  -runner.input.path input.json \
  -runner.output.path output.json \
  -solve.duration 10s \
  -solve.verbosity high \
  -solve.mip.gap.absolute 100 \
  -solve.mip.gap.relative 0.1 \
  -solve.control.float "XPRS_BIGM=10000" \
  -solve.contro.int "XPRS_BREADTHFIRST=7"
```


.golangci.yml
```diff
@@ -333,3 +333,10 @@ issues:
     - path: mip/options\.go
       linters:
         - lll
+    # Although the name stutters, we need to call it mip.MipOptions because the
+    # mip package can also handle linear programs, so potentially we may have
+    # mip.LpOptions.
```
merschformann, line 338: If only we would have named it _linear_ or similar. :sweat_smile: 
For the technically correct: this of course clashes with any quadratic stuff.
Good point though!

mip/options.go
```diff
@@ -1,10 +1,203 @@
...
+	// Gap stopping criteria.
+	Gap GapOptions `json:"gap" usage:"Gap stopping criteria."`
+}
+
+// GapOptions specifies the gap stopping criteria.
+type GapOptions struct {
+	// Absolute gap.
+	Absolute float64 `json:"absolute" usage:"Absolute gap stopping value. If the problem is an integer problem the solver will stop if the gap between the relaxed problem and the best found integer problem is less than this value." default:"0.0"`
+	// Relative gap.
+	Relative float64 `json:"relative" usage:"Relative gap stopping value. If the problem is an integer problem the solver will stop if the relative gap between the relaxed problem and the best found integer problem is less than this value." default:"0.0"`
```
merschformann: Can we survive on the `0.0`? I think we ran into problems with this before due to numerical issues. E.g., Gurobi uses `1e-4` and `1e-10` at default.

templates/legacy/mip-incentive-allocation/main.go
```diff
@@ -108,41 +136,7 @@ func solver(_ context.Context, input input, options options) (schema.Output, err
...
-		return schema.Output{}, err
-	}
-
-	// Set the relative gap to 0% (highs' default is 5%)
-	if err = solveOptions.SetMIPGapRelative(0); err != nil {
-		return schema.Output{}, err
-	}
-
-	// Set verbose level to see a more detailed output
-	solveOptions.SetVerbosity(mip.Off)
```
merschformann, line 132: Where are options now set? Am I missing something obvious? :thinking:

merschformann: Oh, I get it now. It lives in the solver routine. Nevermind. :relaxed:

mip/solver.go
```diff
@@ -1,100 +1,13 @@
...
+// Solver for a MIP problem.
 type Solver interface {
 	// Solve is the entrypoint to solve the model associated with
 	// the invoking solver. Returns a solution when the invoking solver
 	// reaches a conclusion.
 	Solve(options SolveOptions) (Solution, error)
 }
 
-// SolveOptions interface to options for back-end solver.
-type SolveOptions interface {
```
merschformann, line 24: I remember that we back in the day aimed to support options that we expected to be available on all solvers via common functionality. It appears that we are dropping that. Are we aligned about this? Also, this is breaking, no? :thinking: 
I could be missing a bigger point though. :relaxed: 
cc @davidrijsman @dirkschumacher

merschformann (approved):
I have some questions but am generally fond of this. :relaxed:

mip/options.go
sebastian-quintero: The templates had the gap set to 0, so I kept that. To your point, I think let's change it to something similar to your suggestion. Thanks.

merschformann: There was some reason for adapting them in the templates, I think.
But yeah, 0,0 seems pretty aggressive for defaults. :sweat_smile: 
Thank you! :hugs:

mip/solver.go
sebastian-quintero: That is not dropped. The options that are "generally" available were the solver duration, verbosity, absolute gap and relative gap. Those 4 options were the only things you could control. The custom ones had an API but it wasn't connected to anything so you couldn't configure anything else. To preserve that "generally available" characteristic, I kept those 4 options as the ones you could set with the CLI for any provider. I honestly was too lazy to compare highs and xpress and figure out more things they have in common. Lastly, the custom params for each provider can now be set and they work because they are now connected. You just need to know the name, type and the value you want to set.

mip/options.go
```diff
@@ -1,10 +1,203 @@
...
+// SolveOptions are options that can be cionfigured for any solver.
+type SolveOptions struct {
 	// Duration is the maximum duration of the solver. A duration limit of 0 is
 	// treated as infinity.
-	Duration time.Duration `json:"duration" usage:"maximum duration of the solver. A duration limit of 0 is treated as infinity" default:"30s"`
+	Duration time.Duration `json:"duration" usage:"Maximum duration of the solver." default:"30s"`
+	// Verbosity of the solver in the console.
+	Verbosity Verbosity `json:"verbosity" usage:"{off, low, medium, high} Verbosity of the solver in the console." default:"off"`
+	// Mip-specific options.
+	Mip MipOptions `json:"mip" usage:"Options specific to MIP problems. Linear problems do not use these options."`
```
davidrijsman: To me it would make more sense if acronyms are all capital, so

` MIP MIPOptions`

As a side not, these MIPOptions are control options we expose as _hard coded_ options. I would not make a sub-category for the option bening a MIP option. I am pretty sure the users will know what a GAP is (if they touch this option).

mip/options.go
```diff
@@ -1,10 +1,203 @@
...
+	Medium Verbosity = "medium"
+	// High logs everything the underlying logs,
+	// depends on the back-end solver.
+	High Verbosity = "high"
+)
+
+// ControlOptions allow the user to define solver-specific parameters. The
+// parameters' names and types must be known beforehand so that the correct
+// type is used.
+type ControlOptions struct {
```
davidrijsman, line 58: Not sure I completely see how this is working. For CPLEX and other solver the keys of the control options are integers and not strings. Is this implementation taking this into account we might have to support int to {float,int,bool,string} options?

davidrijsman (changes requested):

mip/options.go
sebastian-quintero: The reason for this is so the option is read as `solve.mip.gap.absolute` instead of `solve.gap.absolute`. All the solvers call it a version of MIP gap so I don't want to depart from that.

mip/options.go
sebastian-quintero: CPLEX works like Xpress.
- In Xpress take the [XPRSsetdblcontrol](https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/HTML/XPRSsetdblcontrol.html) function for example. The control parameter is an int and the value of course is a double.
- In CPLEX take the [CPXsetdblparam](https://www.ibm.com/docs/en/icos/22.1.1?topic=s-cpxxsetdblparam-cpxsetdblparam) function, in which the the parameter is also an int and the value is a double.

How I am handling this is I have a map that maps a string to the correct int option. This is part of the Platform PR. For any solver, the idea is to set the param name as a string and the value is set under the correct category.

Notice in the description that this works: `-solve.control.float "XPRS_BIGM=10000"`. Under the hood, `XPRS_BIGM` is translated to an int and the `XPRSsetdblcontrol` function is used as `XPRSsetdblcontrol(problem, 7018, 10000)`.

mip/options.go
sebastian-quintero: I went with what HiGHS uses:
- 1e-6 for absolute
- 1e-4 for relative

mip/options.go
sebastian-quintero: Changed it to MIP though.

davidrijsman (approved):

PR #374, opened 2023-09-14
updates default for min stops to 1

nmisek:
The min stops objective was previously defaulting to 0 so when passing in min_stops and min_stops_penalty in the input, we weren't activating this term in the objective. This sets the default value to 1 so it will always activate when passed through the input.


merschformann (approved):
:rocket:

merschformann (approved):

nilsbeck (approved):

davidrijsman (approved):

PR #376, opened 2023-09-18
ENG-3884 Xpress template

sebastian-quintero:
# Description

Add the `xpress` template demonstrating how to use the Xpress Python interface.

## Changes

* Minor README alignment.
* Minor `ortools` template alignment to include a newline after the docstring of a function.
* Minor adjustment in the `ortools` template to get the solver time from the solver object itself.
* Translate the status coming from a Python solver to `str`.
* Adjusts the `item_id` across MIP templates to be `id`.


templates/ortools/README.md
```diff
@@ -23,15 +23,15 @@ The most important files created are `main.py` and `input.json`.
 Run the command below to check that everything works as expected:
 
 ```bash
-python3 main.py -input input.json -output output.json -duration 30
+python3 main.py -input input.json -duration 30
```
merschformann: In our SDK templates we use the output file. I personally prefer this one that is writing directly to stdout, but should we harmonize the other templates as well? :thinking:

sebastian-quintero: The reason I changed this is because the line below reads:

> The optimal knapsack solution should be printed to stdout.

So we either:
- Save to `output.json` and delete the line below (probably best to be aligned with the other templates).
- Keep it as is to be aligned with what the README states.

merschformann: Oh yeah, that could have been me not updating the README when changing the line around all the time. :grimacing: 

Feel free to make a call. And good catch aligning this. I don't have a strong opinion whether writing to stdout or output.json is preferable.

sebastian-quintero: I wil change it to write to `output.json` per your suggestion.

templates/xpress/main.py
```diff
@@ -0,0 +1,126 @@
...
+        if problem.getSolution(item["variable"]) > 0.9:
+            chosen_items.append(item["item"])
+
+    # Creates the statistics.
+    statistics = {
+        "result": {
+            "custom": {
+                "constraints": problem.getAttrib("rows"),
+                "provider": "xpress",
+                "status": status,
```
merschformann: I compared our HiGHS solver against SCIP via OR-Tools and this clashed a bit. It's probably fine, but we could think about translating this to the same type or even enum (`unbounded`, `optimal`, etc.). The biggest problem was that ortools returns an int here and we return a string like mentioned before. I am not sure what xpress python uses.
It's probably too complicated, but maybe we should at least make them all strings so that experiments aren't confused by the column? :shrug: 
Just something I noticed.

templates/xpress/main.py
```diff
@@ -0,0 +1,126 @@
...
+    problem.setControl("timelimit", duration)
+
+    # Initializes the linear sums.
+    weights = 0.0
+    values = 0.0
+
+    # Creates the decision variables and adds them to the linear sums.
+    items = []
+    for item in input_data["items"]:
+        item_variable = xp.var(vartype=xp.binary, name=item["item_id"])
```
merschformann: Is it too late to change the inputs of the knapsack models from `items.item_id` (stutters) to `items.id`? If it bugs you like me, go for it. However, I also didn't take the time, so, feel free to leave it.
Thanks for keeping them in one format though! :hugs:

templates/xpress/main.py
```diff
@@ -0,0 +1,126 @@
...
+    # Sets the objective function: maximize the value of the chosen items.
+    problem.setObjective(values, sense=xp.maximize)
+
+    # Solves the problem.
+    _, status = problem.optimize()
+
+    # Determines which items were chosen.
+    chosen_items = []
+    for item in items:
+        if problem.getSolution(item["variable"]) > 0.9:
```
merschformann, line 82: After all these years writing MIP models, this `> 0.9` or `> 0.5` or whatever still freaks me out a bit. :sweat_smile:

merschformann (approved):
Awesome! Thank you! :hugs:

templates/xpress/main.py
sebastian-quintero: I'll align with our SDK to print a string.

templates/xpress/main.py
sebastian-quintero: I know, I wanted to set it to `== 1` but @dirkschumacher suggested against it because of the usual suspects. So I left it as is.

templates/xpress/main.py
sebastian-quintero: Consider it done.

templates/xpress/main.py
sebastian-quintero: I now aligned the `main.go` from `mip` to conform to the same style as these.

merschformann: Yeah, it's typical to write it like this. I always used 0.5, but tomato/tomato.
And yeah, always seems a little weird to me too. :sweat_smile:

merschformann (approved):
:heart:

PR #378, opened 2023-09-20
feature/eng-3827-replace-all-nextroute-engine-panics-with-useful-errors

nilsbeck:
# Description

This PR adds an input validation for lat/lon and a couple of other fields.

# Changes

* `input.go`


nextroute/schema/input.go
```diff
@@ -31,54 +31,54 @@ type VehicleDefaults struct {
...
-	MaxStops                *int       `json:"max_stops,omitempty"`
-	MaxDistance             *int       `json:"max_distance,omitempty"`
-	MaxDuration             *int       `json:"max_duration,omitempty"`
-	MaxWait                 *int       `json:"max_wait,omitempty"`
-	CompatibilityAttributes *[]string  `json:"compatibility_attributes,omitempty"`
-	ActivationPenalty       *int       `json:"activation_penalty,omitempty"`
-	AlternateStops          *[]string  `json:"alternate_stops,omitempty"`
+	MinStops                *int       `json:"min_stops,omitempty" minimum:"0"`
+	MinStopsPenalty         *float64   `json:"min_stops_penalty,omitempty" minimum:"0"`
+	MaxStops                *int       `json:"max_stops,omitempty" minimumExclusive:"0"`
```
davidrijsman: This can be zero

nextroute/schema/input.go
```diff
@@ -31,54 +31,54 @@ type VehicleDefaults struct {
...
+	MaxDistance             *int       `json:"max_distance,omitempty" minimumExclusive:"0"`
```
davidrijsman: This can be zero

nextroute/schema/input.go
```diff
@@ -31,54 +31,54 @@ type VehicleDefaults struct {
...
+	MaxDuration             *int       `json:"max_duration,omitempty" minimumExclusive:"0"`
```
davidrijsman: This can be zero

nextroute/schema/input.go
```diff
@@ -31,54 +31,54 @@ type VehicleDefaults struct {
...
+	MaxWait                 *int       `json:"max_wait,omitempty" minimum:"0"`
```
davidrijsman: This can be zero

nextroute/schema/input.go
```diff
@@ -31,54 +31,54 @@ type VehicleDefaults struct {
...
+	CompatibilityAttributes *[]string  `json:"compatibility_attributes,omitempty" uniqueItems:"true"`
+	ActivationPenalty       *int       `json:"activation_penalty,omitempty" minimum:"0"`
```
davidrijsman: This can be zero

nextroute/schema/input.go
```diff
@@ -31,54 +31,54 @@ type VehicleDefaults struct {
...
 
 // Vehicle represents a vehicle.
 type Vehicle struct {
 	Capacity                any            `json:"capacity,omitempty"`
 	StartLevel              any            `json:"start_level,omitempty"`
 	CustomData              any            `json:"custom_data,omitempty"`
-	CompatibilityAttributes *[]string      `json:"compatibility_attributes,omitempty"`
-	MaxDistance             *int           `json:"max_distance,omitempty"`
+	CompatibilityAttributes *[]string      `json:"compatibility_attributes,omitempty" uniqueItems:"true"`
+	MaxDistance             *int           `json:"max_distance,omitempty" minimumExclusive:"0"`
```
davidrijsman: This can be zero

nextroute/schema/input.go
```diff
@@ -31,54 +31,54 @@ type VehicleDefaults struct {
...
-	MinStops                *int           `json:"min_stops,omitempty"`
-	MinStopsPenalty         *float64       `json:"min_stops_penalty,omitempty"`
-	MaxStops                *int           `json:"max_stops,omitempty"`
-	Speed                   *float64       `json:"speed,omitempty"`
-	MaxDuration             *int           `json:"max_duration,omitempty"`
-	MaxWait                 *int           `json:"max_wait,omitempty"`
-	ActivationPenalty       *int           `json:"activation_penalty,omitempty"`
+	MinStops                *int           `json:"min_stops,omitempty" minimum:"0"`
+	MinStopsPenalty         *float64       `json:"min_stops_penalty,omitempty" minimum:"0"`
+	MaxStops                *int           `json:"max_stops,omitempty" minimumExclusive:"0"`
```
davidrijsman: This can be zero

nextroute/schema/input.go
```diff
@@ -31,54 +31,54 @@ type VehicleDefaults struct {
...
+	Speed                   *float64       `json:"speed,omitempty" minimumExclusive:"0"`
+	MaxDuration             *int           `json:"max_duration,omitempty" minimumExclusive:"0"`
```
davidrijsman: This can be zero

davidrijsman (changes requested):
Zero values should be allowed in all marked cases

nextroute/schema/input.go
nilsbeck: No, if set to 0 we throw an error in the construction of expressions: https://github.com/nextmv-io/platform/blob/2d7a22f5c531adab93c11e146e2d52c3bc3adbf1/core/plugins/nextroute/factory/constraint_max_stops.go#L31

nextroute/schema/input.go
nilsbeck: it can't be 0: https://github.com/nextmv-io/platform/blob/2d7a22f5c531adab93c11e146e2d52c3bc3adbf1/core/plugins/nextroute/factory/constraint_distance_limit.go#L37

nextroute/schema/input.go
nilsbeck: it can't be zero: https://github.com/nextmv-io/platform/blob/2d7a22f5c531adab93c11e146e2d52c3bc3adbf1/core/plugins/nextroute/factory/constraint_max_duration.go#L33

nextroute/schema/input.go
nilsbeck: minimum allow the value to be 0, yes

nextroute/schema/input.go
nilsbeck: yes, a minimum of 0 allows the 0 value.

nextroute/schema/input.go
nilsbeck: this cannot be 0: https://github.com/nextmv-io/platform/blob/2d7a22f5c531adab93c11e146e2d52c3bc3adbf1/core/plugins/nextroute/factory/constraint_distance_limit.go#L37

nextroute/schema/input.go
nilsbeck: this cannot be zero: https://github.com/nextmv-io/platform/blob/2d7a22f5c531adab93c11e146e2d52c3bc3adbf1/core/plugins/nextroute/factory/constraint_max_stops.go#L31

nextroute/schema/input.go
nilsbeck: cannot be 0: https://github.com/nextmv-io/platform/blob/2d7a22f5c531adab93c11e146e2d52c3bc3adbf1/core/plugins/nextroute/factory/constraint_max_duration.go#L33

nilsbeck:
> Zero values should be allowed in all marked cases

I added links to where we are throwing an error when allowing the 0 values. so I think this is correct here.

nextroute/schema/input.go
davidrijsman: We should not throw an error when passing a perfectly feasible limit....

nextroute/schema/input.go
davidrijsman: We should not throw an error when passing a perfectly feasible limit....

nextroute/schema/input.go
davidrijsman: We should not throw an error when passing a perfectly feasible limit....

nextroute/schema/input.go
davidrijsman: We should not throw an error when passing a perfectly feasible limit....

nextroute/schema/input.go
davidrijsman: We should not throw an error when passing a perfectly feasible limit....

nextroute/schema/input.go
davidrijsman: We should not throw an error when passing a perfectly feasible limit....

davidrijsman (changes requested):
We should not compound incorrect behaviour, I would fix the incorrect errors

nilsbeck:
> We should not compound incorrect behaviour, I would fix the incorrect errors

I am not sure that this is a good approach. I forsee that this will create work for us because customers accidentally will pass "0" for max stop or max distance e.g. and complain that those vehicles are not assigned anything.
While the other way around, if we reject "0" for those fields they will find a way to get what they actually want without our help.
I'd like to get some broader feedback on my thought here: @sebastian-quintero @merschformann @dirkschumacher @muellerd (added Daniel here as well because he does the most customer work right now)

sebastian-quintero:
> > We should not compound incorrect behaviour, I would fix the incorrect errors
> 
> I am not sure that this is a good approach. I forsee that this will create work for us because customers accidentally will pass "0" for max stop or max distance e.g. and complain that those vehicles are not assigned anything. While the other way around, if we reject "0" for those fields they will find a way to get what they actually want without our help. I'd like to get some broader feedback on my thought here: @sebastian-quintero @merschformann @dirkschumacher @muellerd (added Daniel here as well because he does the most customer work right now)

I see the point for both really:
1. You are strict with what you allow and the user has to make sure that the input is always correct. Passing in a vehicle with max distance set to zero will give an error because that doesn't make sense. The user corrects the inputs.
1. You are flexible and you let the engine figure out what to do. A vehicle with max distance set to 0 cannot be assigned anything. The engine does not use that vehicle and simply assigns stops to other vehicles. This does not give an error because it is always assuming that the input is correct, this is, there is intention behind setting max distance to 0.

I think the more flexible we are, the better. If for performance reasons we would like to be stringent and not allow "incorrect" settings, then let's go with option 1 and validate the semantics of the input thoroughly. However, production systems often benefit from going with 2 because in case some service fails and sends 0s on some settings, you guarantee some service. That begs the question though: what is better? No solution or a bad solution? 

There is no definitive answer of course so my actual preference is: let's make it an option. You know? Something like `-flexibility {off, on}` or whatever. When on, there are no zeroes allowed. When off, we make sure that even `nextroute` can tolerate a zero being passed as a limit.

Default behavior? 🤷🏻‍♂️  up for debate.

davidrijsman:
For the float64 values you would not allow 0.0 but would 0.001, totally arbitrary. My view remains that an error ony if it is incorrect input not when we believe the user did not want to do this.

davidrijsman (approved):

PR #383, opened 2023-09-27
Feature/eng 3430 add comments to inputoutput structs

davidrijsman:
Document input and output schema of next route.

### DIFF
--- a/nextroute/schema/input.go
+++ b/nextroute/schema/input.go
@@ -5,7 +5,7 @@ import (
 	"time"
 )
 
-// Input is the default input schema for next route.
+// Input is the default input schema for nextroute.
 type Input struct {
 	// Options arbitrary options.
 	Options any `json:"options,omitempty"`
--- a/nextroute/schema/input.go
+++ b/nextroute/schema/input.go
@@ -5,7 +5,7 @@ import (
 	"time"
 )
 
-// Input is the default input schema for nextroute.
+// Input is the default input schema for next route.
 type Input struct {
 	// Options arbitrary options.
 	Options any `json:"options,omitempty"`
--- a/nextroute/schema/input.go
+++ b/nextroute/schema/input.go
@@ -7,129 +7,217 @@ import (
 
 // Input is the default input schema for nextroute.
 type Input struct {
-	Options        any              `json:"options,omitempty"`
-	CustomData     any              `json:"custom_data,omitempty"`
-	Defaults       *Defaults        `json:"defaults,omitempty"`
-	StopGroups     *[][]string      `json:"stop_groups,omitempty"`
-	DurationMatrix *[][]float64     `json:"duration_matrix,omitempty"`
-	DistanceMatrix *[][]float64     `json:"distance_matrix,omitempty"`
+	// Options arbitrary options.
+	Options any `json:"options,omitempty"`
+	// CustomData arbitrary custom data.
+	CustomData any `json:"custom_data,omitempty"`
+	// Defaults default properties for vehicles and stops.
+	Defaults *Defaults `json:"defaults,omitempty"`
+	// StopGroups group of stops that must be part of the same route.
+	StopGroups *[][]string `json:"stop_groups,omitempty"`
+	// DurationMatrix matrix of durations in seconds between stops.
+	DurationMatrix *[][]float64 `json:"duration_matrix,omitempty"`
+	// DistanceMatrix matrix of distances in meters between stops.
+	DistanceMatrix *[][]float64 `json:"distance_matrix,omitempty"`
+	// DurationGroups duration in seconds added when approaching the group.
 	DurationGroups *[]DurationGroup `json:"duration_groups,omitempty"`
-	Vehicles       []Vehicle        `json:"vehicles,omitempty"`
-	Stops          []Stop           `json:"stops,omitempty"`
+	// Vehicles to route.
+	Vehicles []Vehicle `json:"vehicles,omitempty"`
+	// Stops that will be routed and assigned to the vehicles.
+	Stops []Stop `json:"stops,omitempty"`
+	// AlternateStops a set of alternate stops for vehicles.
 	AlternateStops *[]AlternateStop `json:"alternate_stops,omitempty"`
 }
 
 // Defaults contains default values for vehicles and stops.
 type Defaults struct {
+	// Vehicles default values for vehicles.
 	Vehicles *VehicleDefaults `json:"vehicles,omitempty"`
-	Stops    *StopDefaults    `json:"stops,omitempty"`
+	// Stops default values for stops.
+	Stops *StopDefaults `json:"stops,omitempty"`
 }
 
 // VehicleDefaults contains default values for vehicles.
 type VehicleDefaults struct {
-	Capacity                any        `json:"capacity,omitempty"`
-	StartLevel              any        `json:"start_level,omitempty"`
-	StartLocation           *Location  `json:"start_location,omitempty"`
-	EndLocation             *Location  `json:"end_location,omitempty"`
-	Speed                   *float64   `json:"speed,omitempty"`
-	StartTime               *time.Time `json:"start_time,omitempty"`
-	EndTime                 *time.Time `json:"end_time,omitempty"`
-	MinStops                *int       `json:"min_stops,omitempty"`
-	MinStopsPenalty         *float64   `json:"min_stops_penalty,omitempty"`
-	MaxStops                *int       `json:"max_stops,omitempty"`
-	MaxDistance             *int       `json:"max_distance,omitempty"`
-	MaxDuration             *int       `json:"max_duration,omitempty"`
-	MaxWait                 *int       `json:"max_wait,omitempty"`
-	CompatibilityAttributes *[]string  `json:"compatibility_attributes,omitempty"`
-	ActivationPenalty       *int       `json:"activation_penalty,omitempty"`
-	AlternateStops          *[]string  `json:"alternate_stops,omitempty"`
+	// Capacity of the vehicle.
+	Capacity any `json:"capacity,omitempty"`
+	// StartLevel initial level of the vehicle.
+	StartLevel any `json:"start_level,omitempty"`
+	// StartLocation location where the vehicle starts..
+	StartLocation *Location `json:"start_location,omitempty"`
+	// EndLocation location where the vehicle ends..
+	EndLocation *Location `json:"end_location,omitempty"`
+	// Speed of the vehicle in meters per second.
+	Speed *float64 `json:"speed,omitempty"`
+	// StartTime time when the vehicle starts its route.
+	StartTime *time.Time `json:"start_time,omitempty"`
+	// EndTime latest time at which the vehicle ends its route.
+	EndTime *time.Time `json:"end_time,omitempty"`
+	// MinStops minimum stops that a vehicle should visit.
+	MinStops *int `json:"min_stops,omitempty"`
+	// MinStopsPenalty penalty for not visiting the minimum number of stops.
+	MinStopsPenalty *float64 `json:"min_stops_penalty,omitempty"`
+	// MaxStops maximum number of stops that the vehicle can visit.
+	MaxStops *int `json:"max_stops,omitempty"`
+	// MaxDistance maximum distance in meters that the vehicle can travel.
+	MaxDistance *int `json:"max_distance,omitempty"`
+	// MaxDuration maximum duration in seconds that the vehicle can travel.
+	MaxDuration *int `json:"max_duration,omitempty"`
+	// MaxWait maximum aggregated waiting time that the vehicle can wait across route stops.
+	MaxWait *int `json:"max_wait,omitempty"`
+	// CompatibilityAttributes attributes that the vehicle is compatible with.
+	CompatibilityAttributes *[]string `json:"compatibility_attributes,omitempty"`
+	// ActivationPenalty penalty of using the vehicle.
+	ActivationPenalty *int `json:"activation_penalty,omitempty"`
+	// AlternateStops a set of alternate stops for which only one should be serviced.
+	AlternateStops *[]string `json:"alternate_stops,omitempty"`
 }
 
 // StopDefaults contains default values for stops.
 type StopDefaults struct {
-	UnplannedPenalty        *int       `json:"unplanned_penalty,omitempty"`
-	Quantity                any        `json:"quantity,omitempty"`
-	StartTimeWindow         any        `json:"start_time_window,omitempty"`
-	MaxWait                 *int       `json:"max_wait,omitempty"`
-	Duration                *int       `json:"duration,omitempty"`
-	TargetArrivalTime       *time.Time `json:"target_arrival_time,omitempty"`
-	EarlyArrivalTimePenalty *float64   `json:"early_arrival_time_penalty,omitempty"`
-	LateArrivalTimePenalty  *float64   `json:"late_arrival_time_penalty,omitempty"`
-	CompatibilityAttributes *[]string  `json:"compatibility_attributes,omitempty"`
+	// UnplannedPenalty penalty for not planning a stop.
+	UnplannedPenalty *int `json:"unplanned_penalty,omitempty"`
+	// Quantity of the stop.
+	Quantity any `json:"quantity,omitempty"`
+	// StartTimeWindow time window in which the stop can start service.
+	StartTimeWindow any `json:"start_time_window,omitempty"`
+	// MaxWait maximum waiting duration in seconds at the stop.
+	MaxWait *int `json:"max_wait,omitempty"`
+	// Duration in seconds that the stop takes.
+	Duration *int `json:"duration,omitempty"`
+	// TargetArrivalTime at the stop.
+	TargetArrivalTime *time.Time `json:"target_arrival_time,omitempty"`
+	// EarlyArrivalTimePenalty penalty per second for arriving at the stop before the target arrival time.
+	EarlyArrivalTimePenalty *float64 `json:"early_arrival_time_penalty,omitempty"`
+	// LateArrivalTimePenalty penalty per second for arriving at the stop after the target arrival time.
+	LateArrivalTimePenalty *float64 `json:"late_arrival_time_penalty,omitempty"`
+	// CompatibilityAttributes attributes that the stop is compatible with.
+	CompatibilityAttributes *[]string `json:"compatibility_attributes,omitempty"`
 }
 
 // Vehicle represents a vehicle.
 type Vehicle struct {
-	Capacity                any            `json:"capacity,omitempty"`
-	StartLevel              any            `json:"start_level,omitempty"`
-	CustomData              any            `json:"custom_data,omitempty"`
-	CompatibilityAttributes *[]string      `json:"compatibility_attributes,omitempty"`
-	MaxDistance             *int           `json:"max_distance,omitempty"`
-	StopDurationMultiplier  *float64       `json:"stop_duration_multiplier,omitempty"`
-	StartTime               *time.Time     `json:"start_time,omitempty"`
-	EndTime                 *time.Time     `json:"end_time,omitempty"`
-	EndLocation             *Location      `json:"end_location,omitempty"`
-	MinStops                *int           `json:"min_stops,omitempty"`
-	MinStopsPenalty         *float64       `json:"min_stops_penalty,omitempty"`
-	MaxStops                *int           `json:"max_stops,omitempty"`
-	Speed                   *float64       `json:"speed,omitempty"`
-	MaxDuration             *int           `json:"max_duration,omitempty"`
-	MaxWait                 *int           `json:"max_wait,omitempty"`
-	ActivationPenalty       *int           `json:"activation_penalty,omitempty"`
-	StartLocation           *Location      `json:"start_location,omitempty"`
-	AlternateStops          *[]string      `json:"alternate_stops,omitempty"`
-	InitialStops            *[]InitialStop `json:"initial_stops,omitempty"`
-	ID                      string         `json:"id,omitempty"`
+	// Capacity of the vehicle.
+	Capacity any `json:"capacity,omitempty"`
+	// StartLevel initial level of the vehicle.
+	StartLevel any `json:"start_level,omitempty"`
+	// CustomData arbitrary custom data.
+	CustomData any `json:"custom_data,omitempty"`
+	// CompatibilityAttributes attributes that the vehicle is compatible with.
+	CompatibilityAttributes *[]string `json:"compatibility_attributes,omitempty"`
+	// MaxDistance maximum distance in meters that the vehicle can travel.
+	MaxDistance *int `json:"max_distance,omitempty"`
+	// StopDurationMultiplier multiplier for the duration of stops.
+	StopDurationMultiplier *float64 `json:"stop_duration_multiplier,omitempty"`
+	// StartTime time when the vehicle starts its route.
+	StartTime *time.Time `json:"start_time,omitempty"`
+	// EndTime latest time at which the vehicle ends its route.
+	EndTime *time.Time `json:"end_time,omitempty"`
+	// EndLocation location where the vehicle ends.
+	EndLocation *Location `json:"end_location,omitempty"`
+	// MinStops minimum stops that a vehicle should visit.
+	MinStops *int `json:"min_stops,omitempty"`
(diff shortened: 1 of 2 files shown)


nextroute/schema/input.go
```diff
@@ -5,131 +5,219 @@ import (
 	"time"
 )
 
-// Input is the default input schema for nextroute.
+// Input is the default input schema for next route.
```
dirkschumacher: next route or nextroute?

dirkschumacher (approved):

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
 // Input is the default input schema for nextroute.
 type Input struct {
-	Options        any              `json:"options,omitempty"`
-	CustomData     any              `json:"custom_data,omitempty"`
-	Defaults       *Defaults        `json:"defaults,omitempty"`
-	StopGroups     *[][]string      `json:"stop_groups,omitempty"`
-	DurationMatrix *[][]float64     `json:"duration_matrix,omitempty"`
-	DistanceMatrix *[][]float64     `json:"distance_matrix,omitempty"`
+	// Options arbitrary options.
+	Options any `json:"options,omitempty"`
```
merschformann, line 11: Do we still need these anymore? We should probably add an issue for this.

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
+	// CustomData arbitrary custom data.
```
merschformann, line 12: I think we should explain here (or touch a bit on it) what this is for. I.e., this is for the customer to store custom data on the (in this case the input) struct.

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
+	CustomData any `json:"custom_data,omitempty"`
+	// Defaults default properties for vehicles and stops.
```
merschformann, line 14: Another sentence like this would be helpful to me:
"Default values are applied to stops and vehicles that do not set them explicitly."

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
+	Defaults *Defaults `json:"defaults,omitempty"`
+	// StopGroups group of stops that must be part of the same route.
+	StopGroups *[][]string `json:"stop_groups,omitempty"`
+	// DurationMatrix matrix of durations in seconds between stops.
```
merschformann, line 18: I think we should also add:
"The indices of the matrix are defined as m[i][j] with i as 'from' and j as 'to'. Furthermore, the range of matrix first spans all stops of the input (same order) and then start/end locations of the vehicles in vehicle order. E.g.: [stop_1, stop_2, ..., stop_n, start_1, end_1, ..., start_m, end_m]"
Or something like this. Alternatively, we can link the docs, if we have something suitable to link to instead.

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
+	DurationMatrix *[][]float64 `json:"duration_matrix,omitempty"`
+	// DistanceMatrix matrix of distances in meters between stops.
```
merschformann, line 20: Same comment as DurationMatrix. :blush:

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
+	DistanceMatrix *[][]float64 `json:"distance_matrix,omitempty"`
+	// DurationGroups duration in seconds added when approaching the group.
 	DurationGroups *[]DurationGroup `json:"duration_groups,omitempty"`
-	Vehicles       []Vehicle        `json:"vehicles,omitempty"`
-	Stops          []Stop           `json:"stops,omitempty"`
+	// Vehicles to route.
+	Vehicles []Vehicle `json:"vehicles,omitempty"`
+	// Stops that will be routed and assigned to the vehicles.
+	Stops []Stop `json:"stops,omitempty"`
+	// AlternateStops a set of alternate stops for vehicles.
```
merschformann, line 28: We should explain in some more detail how alternate stops will be handled by the solver. In short. Details are for docs. :relaxed:

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
-	MaxDuration             *int       `json:"max_duration,omitempty"`
-	MaxWait                 *int       `json:"max_wait,omitempty"`
-	CompatibilityAttributes *[]string  `json:"compatibility_attributes,omitempty"`
-	ActivationPenalty       *int       `json:"activation_penalty,omitempty"`
-	AlternateStops          *[]string  `json:"alternate_stops,omitempty"`
+	// Capacity of the vehicle.
+	Capacity any `json:"capacity,omitempty"`
+	// StartLevel initial level of the vehicle.
+	StartLevel any `json:"start_level,omitempty"`
+	// StartLocation location where the vehicle starts..
```
merschformann, line 46:
```suggestion
	// StartLocation location where the vehicle starts.
```
I would add something like: "If omitted, the vehicle starts at the first stop it services."

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
+	StartLocation *Location `json:"start_location,omitempty"`
+	// EndLocation location where the vehicle ends..
+	EndLocation *Location `json:"end_location,omitempty"`
+	// Speed of the vehicle in meters per second.
```
merschformann, line 50: We need to add that this is not used when adding a `duration_matrix`, but will be used in conjunction with the `distance_matrix`.

nilsbeck:
we may also want to use the "usage" tag to do those explanations in addition to a comment above the field. we can also use the field "doc" which we enabled because we use the `huma` library

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
+	Speed *float64 `json:"speed,omitempty"`
+	// StartTime time when the vehicle starts its route.
```
merschformann, line 52: Is this optional at the moment? Should we add that it is mandatory? Or explain how it behaves when it is omitted?

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
+	StartTime *time.Time `json:"start_time,omitempty"`
+	// EndTime latest time at which the vehicle ends its route.
+	EndTime *time.Time `json:"end_time,omitempty"`
+	// MinStops minimum stops that a vehicle should visit.
```
merschformann, line 56: We need to explain that this is a soft-constraint, especially, considering that `max_stops` is defined right around the corner. Maybe something like: "This is a soft constraint. I.e., if a vehicle in terms of stops serviced falls short of this number, the `MinStopsPenalty` will be applied."

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
+	MinStops *int `json:"min_stops,omitempty"`
+	// MinStopsPenalty penalty for not visiting the minimum number of stops.
+	MinStopsPenalty *float64 `json:"min_stops_penalty,omitempty"`
+	// MaxStops maximum number of stops that the vehicle can visit.
```
merschformann, line 60: We need to clarify that this is a hard constraint. Something like: "This is a hard constraint. I.e., the solver will never allow the vehicle to service more stops than this value."

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
+	MaxStops *int `json:"max_stops,omitempty"`
+	// MaxDistance maximum distance in meters that the vehicle can travel.
```
merschformann, line 62: A short note that this is hard constraint is probably wise, since we don't distinguish hard vs. soft on the field name. This applies for other fields here as well.

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
+	MaxDistance *int `json:"max_distance,omitempty"`
+	// MaxDuration maximum duration in seconds that the vehicle can travel.
+	MaxDuration *int `json:"max_duration,omitempty"`
+	// MaxWait maximum aggregated waiting time that the vehicle can wait across route stops.
```
merschformann, line 66: I would add: "Waiting time is the time a vehicle waits when it arrives early at a stop, i.e., before the stop's `StartTimeWindow` opens.

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
+	MaxWait *int `json:"max_wait,omitempty"`
+	// CompatibilityAttributes attributes that the vehicle is compatible with.
```
merschformann, line 68: I would add: "A vehicle is compatible with a stop, if it has at least one attribute that the stop demands."

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
+	CompatibilityAttributes *[]string `json:"compatibility_attributes,omitempty"`
+	// ActivationPenalty penalty of using the vehicle.
+	ActivationPenalty *int `json:"activation_penalty,omitempty"`
+	// AlternateStops a set of alternate stops for which only one should be serviced.
```
merschformann, line 72: We _could_ add: "These stops are defined in the separate `alternate_stops` array."

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
-	StartTimeWindow         any        `json:"start_time_window,omitempty"`
-	MaxWait                 *int       `json:"max_wait,omitempty"`
-	Duration                *int       `json:"duration,omitempty"`
-	TargetArrivalTime       *time.Time `json:"target_arrival_time,omitempty"`
-	EarlyArrivalTimePenalty *float64   `json:"early_arrival_time_penalty,omitempty"`
-	LateArrivalTimePenalty  *float64   `json:"late_arrival_time_penalty,omitempty"`
-	CompatibilityAttributes *[]string  `json:"compatibility_attributes,omitempty"`
+	// UnplannedPenalty penalty for not planning a stop.
+	UnplannedPenalty *int `json:"unplanned_penalty,omitempty"`
+	// Quantity of the stop.
```
merschformann, line 80: We really should explain how quantities work. E.g., we could add something like this: "A negative (-) quantity consumes available capacity of a vehicle (e.g.: the vehicle picks something up), while a positive (+) quantity replenishes capacity of a vehicle (e.g.: the vehicle drops something off). Quantities and capacities can be scalar or multi-dimensional. I.e., a simple int or float can be used, or a dictionary of different capacity types (strings) to int/float."

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
-	MinStops                *int       `json:"min_stops,omitempty"`
-	MinStopsPenalty         *float64   `json:"min_stops_penalty,omitempty"`
-	MaxStops                *int       `json:"max_stops,omitempty"`
-	MaxDistance             *int       `json:"max_distance,omitempty"`
-	MaxDuration             *int       `json:"max_duration,omitempty"`
-	MaxWait                 *int       `json:"max_wait,omitempty"`
-	CompatibilityAttributes *[]string  `json:"compatibility_attributes,omitempty"`
-	ActivationPenalty       *int       `json:"activation_penalty,omitempty"`
-	AlternateStops          *[]string  `json:"alternate_stops,omitempty"`
+	// Capacity of the vehicle.
```
merschformann, line 42: We should add: "Quantities and capacities can be scalar or multi-dimensional. I.e., a simple int or float can be used, or a dictionary of different capacity types (strings) to int/float."

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
-	Duration                *int       `json:"duration,omitempty"`
-	TargetArrivalTime       *time.Time `json:"target_arrival_time,omitempty"`
-	EarlyArrivalTimePenalty *float64   `json:"early_arrival_time_penalty,omitempty"`
-	LateArrivalTimePenalty  *float64   `json:"late_arrival_time_penalty,omitempty"`
-	CompatibilityAttributes *[]string  `json:"compatibility_attributes,omitempty"`
+	// UnplannedPenalty penalty for not planning a stop.
+	UnplannedPenalty *int `json:"unplanned_penalty,omitempty"`
+	// Quantity of the stop.
+	Quantity any `json:"quantity,omitempty"`
+	// StartTimeWindow time window in which the stop can start service.
```
merschformann, line 82: We need to roughly outline which kind of types can be used here. E.g.: "This can be a single window given as a 2-long array of RFC3339 strings or time.Time (e.g.: `[windowStart, windowEnd]`). Or, an array of multiple of these windows."
I think we allow both (RFC3339 string and time.Time). We should maybe confirm or just choose to use one here (probably time.Time then).

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
+	StartTimeWindow any `json:"start_time_window,omitempty"`
+	// MaxWait maximum waiting duration in seconds at the stop.
```
merschformann, line 84: I would add the same comment as for the vehicle here, just to make sure a user does not miss it when only using the one on the stop (or vice versa).

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
+	MaxWait *int `json:"max_wait,omitempty"`
+	// Duration in seconds that the stop takes.
```
merschformann, line 86:
```suggestion
	// Duration in seconds that the vehicle spends servicing / at the stop.
```
Maybe a little clearer. Feel free to disregard.

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
+	Duration *int `json:"duration,omitempty"`
+	// TargetArrivalTime at the stop.
```
merschformann, line 88: Maybe add: "This is the reference time used by the `EarlyArrivalTimePenalty` and `LateArrivalTimePenalty` penalties."

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
+	TargetArrivalTime *time.Time `json:"target_arrival_time,omitempty"`
+	// EarlyArrivalTimePenalty penalty per second for arriving at the stop before the target arrival time.
+	EarlyArrivalTimePenalty *float64 `json:"early_arrival_time_penalty,omitempty"`
+	// LateArrivalTimePenalty penalty per second for arriving at the stop after the target arrival time.
+	LateArrivalTimePenalty *float64 `json:"late_arrival_time_penalty,omitempty"`
+	// CompatibilityAttributes attributes that the stop is compatible with.
```
merschformann, line 94: Maybe a quick elaboration how the logic works like with the vehicle comment?

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
-	MaxStops                *int       `json:"max_stops,omitempty"`
-	MaxDistance             *int       `json:"max_distance,omitempty"`
-	MaxDuration             *int       `json:"max_duration,omitempty"`
-	MaxWait                 *int       `json:"max_wait,omitempty"`
-	CompatibilityAttributes *[]string  `json:"compatibility_attributes,omitempty"`
-	ActivationPenalty       *int       `json:"activation_penalty,omitempty"`
-	AlternateStops          *[]string  `json:"alternate_stops,omitempty"`
+	// Capacity of the vehicle.
+	Capacity any `json:"capacity,omitempty"`
+	// StartLevel initial level of the vehicle.
```
merschformann, line 44: We need to tie this to the capacity (it isn't obvious). And potentially should remark again that this is also potentially multi-dimensional (or can be used multi-dimensionally). :sweat_smile:

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
+	// EarlyArrivalTimePenalty penalty per second for arriving at the stop before the target arrival time.
+	EarlyArrivalTimePenalty *float64 `json:"early_arrival_time_penalty,omitempty"`
+	// LateArrivalTimePenalty penalty per second for arriving at the stop after the target arrival time.
+	LateArrivalTimePenalty *float64 `json:"late_arrival_time_penalty,omitempty"`
+	// CompatibilityAttributes attributes that the stop is compatible with.
+	CompatibilityAttributes *[]string `json:"compatibility_attributes,omitempty"`
 }
 
 // Vehicle represents a vehicle.
 type Vehicle struct {
```
merschformann, line 99: I would copy the (extended) comments of the defaults to the vehicle itself (where applicable).

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
+	InitialStops *[]InitialStop `json:"initial_stops,omitempty"`
+	// ID of the vehicle.
+	ID string `json:"id,omitempty"`
 }
 
 // InitialStop represents an initial stop.
 type InitialStop struct {
-	Fixed *bool  `json:"fixed,omitempty"`
-	ID    string `json:"id"`
+	// Fixed whether the stop is fixed or not.
```
merschformann, line 144:
```suggestion
	// Fixed ensures that the stop stays with the vehicle it is defined on. Furthermore, the sequence of fixed stops will stay untouched. However, other stops may get assigned in between fixed stops.
```
We should explain what it does instead.

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
+	Fixed *bool `json:"fixed,omitempty"`
+	// ID unique identifier for the stop.
+	ID string `json:"id"`
 }
 
 // AlternateStop represents an alternate stop.
 type AlternateStop struct {
```
merschformann, line 151: We can copy comments from above to this one again to have all the outline infos in the places where they are needed.

nextroute/schema/input.go
```diff
@@ -7,129 +7,217 @@ import (
...
+	// TargetArrivalTime at the stop.
+	TargetArrivalTime *time.Time `json:"target_arrival_time,omitempty"`
+	// ID unique identifier for the stop.
+	ID string `json:"id,omitempty"`
+	// Location where the stop is.
+	Location Location `json:"location,omitempty"`
 }
 
 // Stop represents a stop.
 type Stop struct {
```
merschformann, line 177: Again, we can copy the comments from above to this one to fill in some gaps.

nextroute/schema/output.go
```diff
@@ -6,60 +6,96 @@ import (
...
-	Unplanned []StopOutput    `json:"unplanned"`
-	Vehicles  []VehicleOutput `json:"vehicles"`
+	// Unplanned is the list of stops that were not planned in the solution.
+	Unplanned []StopOutput `json:"unplanned"`
+	// Vehicles is the list of vehicles that were used in the solution.
+	Vehicles []VehicleOutput `json:"vehicles"`
+	// Objective is the objective of the solution.
 	Objective ObjectiveOutput `json:"objective"`
-	Check     *check.Output   `json:"check,omitempty"`
+	// Check is the check of the solution.
```
merschformann, line 17: This sounds confusing. Isn't it more like the result of the check operation?

nextroute/schema/output.go
```diff
@@ -6,60 +6,96 @@ import (
...
 )
 
-// SolutionOutput represents a solutions as JSON.
+// SolutionOutput represents a solution as JSON.
 type SolutionOutput struct {
-	Unplanned []StopOutput    `json:"unplanned"`
-	Vehicles  []VehicleOutput `json:"vehicles"`
+	// Unplanned is the list of stops that were not planned in the solution.
+	Unplanned []StopOutput `json:"unplanned"`
+	// Vehicles is the list of vehicles that were used in the solution.
```
merschformann, line 13: Doesn't this list _all_ vehicles of the input? (not only the ones _used_ in the solution - this makes it sound related to the activation penalty)

nextroute/schema/output.go
```diff
@@ -6,60 +6,96 @@ import (
...
+	Vehicles []VehicleOutput `json:"vehicles"`
+	// Objective is the objective of the solution.
```
merschformann, line 15: Isn't this statistics / information about the solution's objective value? The "objective of the solution" makes it sound like it defines the models objective. :thinking:

nextroute/schema/output.go
```diff
@@ -6,60 +6,96 @@ import (
...
 // StopOutput is the basic struct for a stop.
 type StopOutput struct {
-	ID         string   `json:"id"`
-	Location   Location `json:"location"`
-	CustomData any      `json:"custom_data,omitempty"`
+	// ID is the ID of the stop.
+	ID string `json:"id"`
+	// Location is the location of the stop.
+	Location Location `json:"location"`
+	// CustomData is the custom data of the stop.
```
merschformann, line 27: I would add: "This will contain any custom data set by the user on the input stop." (or similar)

nextroute/schema/output.go
```diff
@@ -6,60 +6,96 @@ import (
...
+	RouteTravelDuration int `json:"route_travel_duration"`
+	// RouteTravelDistance is the total travel distance of the vehicle.
+	RouteTravelDistance int `json:"route_travel_distance,omitempty"`
+	// RouteStopsDuration is the total stops duration of the vehicle.
+	RouteStopsDuration int `json:"route_stops_duration,omitempty"`
+	// RouteWaitingDuration is the total waiting duration of the vehicle.
+	RouteWaitingDuration int `json:"route_waiting_duration,omitempty"`
+	// RouteDuration is the total duration of the vehicle.
+	RouteDuration int `json:"route_duration"`
+	// CustomData is the custom data of the vehicle.
```
merschformann, line 47:
```suggestion
	// CustomData is the custom data of the vehicle as provided by the user.
```
Maybe like this?

nextroute/schema/output.go
```diff
@@ -6,60 +6,96 @@ import (
...
-	WaitingDuration          int        `json:"waiting_duration,omitempty"`
-	StartTime                *time.Time `json:"start_time,omitempty"`
-	Duration                 int        `json:"duration,omitempty"`
-	EndTime                  *time.Time `json:"end_time,omitempty"`
-	EarlyArrivalDuration     int        `json:"early_arrival_duration,omitempty"`
-	LateArrivalDuration      int        `json:"late_arrival_duration,omitempty"`
-	CustomData               any        `json:"custom_data,omitempty"`
+	// Stop is the input stop.
+	Stop StopOutput `json:"stop"`
+	// TravelDuration is the travel duration of the stop.
```
merschformann, line 57:
```suggestion
	// TravelDuration is the travel duration from the previous stop to this stop (in seconds).
```
Maybe this is a bit clearer?

nextroute/schema/output.go
```diff
@@ -6,60 +6,96 @@ import (
...
+	TravelDuration int `json:"travel_duration"`
+	// CumulativeTravelDuration is the total travel duration to get to this location.
+	CumulativeTravelDuration int `json:"cumulative_travel_duration"`
+	// TravelDistance is the travel distance of the stop.
```
merschformann, line 61:
```suggestion
	// TravelDistance is the travel distance from the previous stop to this stop (in meters).
```
Same here? :thinking:

nextroute/schema/output.go
```diff
@@ -6,60 +6,96 @@ import (
...
+	TravelDistance int `json:"travel_distance,omitempty"`
+	// CumulativeTravelDistance is the total travel distance to get to this location.
+	CumulativeTravelDistance int `json:"cumulative_travel_distance,omitempty"`
+	// TargetArrivalTime is the target arrival time of the stop.
+	TargetArrivalTime *time.Time `json:"target_arrival_time,omitempty"`
+	// ArrivalTime is the arrival time of the stop.
+	ArrivalTime *time.Time `json:"arrival_time,omitempty"`
+	// WaitingDuration is the waiting duration of the stop in seconds.
+	WaitingDuration int `json:"waiting_duration,omitempty"`
+	// StartTime is the start time of the stop.
```
merschformann, line 71:
```suggestion
	// StartTime is the time the vehicle starts servicing this stop.
```
Start time sounds a bit imprecise to me (it could be that we count driving towards the stop already as part of starting "work" on the stop or so). I am not sure whether we like servicing though. Maybe there is a better option? :shrug:

nextroute/schema/output.go
```diff
@@ -6,60 +6,96 @@ import (
...
+	StartTime *time.Time `json:"start_time,omitempty"`
+	// Duration is the duration of the stop in seconds.
+	Duration int `json:"duration,omitempty"`
+	// EndTime is the end time of the stop.
```
merschformann, line 75: Same comment as for the StartTime.

merschformann:
Sorry for the comment spam. And the PR was merged a bit too quickly under my nose. The intention behind my request for comments on these structs was to be a bit more detailed about their functionality.

PR #384, opened 2023-09-28
Introduces no-mix constraint

davidrijsman:
Introduces no-mix constraint


merschformann (commented):
Looks good. Still going a bit back and forth on whether we should prepend the `no_` everywhere or just omit it everywhere. Or, use the current mix. It's probably not too important. :sweat_smile:

merschformann (approved):

nextroute/schema/input.go
```diff
@@ -203,6 +203,17 @@ type Stop struct {
 	ID string `json:"id,omitempty"`
 	// Location where the stop is.
 	Location Location `json:"location,omitempty"`
+	// MixingItems defines the items that are inserted or removed from the vehicle when visiting the stop.
+	MixingItems any `json:"mixing_items,omitempty"`
```
sebastian-quintero, line 207: Should this be of type `MixItem`?

nextroute/schema/input.go
```diff
@@ -203,6 +203,17 @@ type Stop struct {
...
+}
+
+// MixItem is an item that is used to specify the type of mix.
+type MixItem struct {
+	// Name is the name of the mix item.
```
sebastian-quintero, line 212:
```suggestion
	// Name of the mix item.
```

nextroute/schema/output.go
```diff
@@ -78,6 +78,8 @@ type PlannedStopOutput struct {
 	EarlyArrivalDuration int `json:"early_arrival_duration,omitempty"`
 	// LateArrivalDuration is the late arrival duration of the stop in seconds.
 	LateArrivalDuration int `json:"late_arrival_duration,omitempty"`
+	// MixItems is the mix items of the stop.
+	MixItems any `json:"mix_items,omitempty"`
```
sebastian-quintero, line 82: Same question regarding the type.

sebastian-quintero (approved):
Should the new `Stop` property be also available in `AlternateStop` and/or `StopDefaults`?

nextroute/schema/input.go
davidrijsman: No there can be multiple items that can not be mixed on a vehicle

nextroute/schema/output.go
davidrijsman: Same answer, there can be multiple items that can not be mixed

PR #386, opened 2023-10-02
Give better support for showing xpress cannot run for Apple silicon macs

sebastian-quintero:
Give better support for showing that Xpress is not installable/callable from an M-X Apple Mac.


merschformann (approved):
:rocket:

PR #392, opened 2023-10-30
Introduce initial solution construction heuristic for stop groups.

davidrijsman:
# Description

Introduces a construction heuristic which can deal with stopgroups.

Can be used as following:

```
solution, err := factory.NewGreedySolution(
		ctx,
		input,
		options.Model,
		sdkNextRouteFactory.GreedySolutionOptions{
			ClusterSolutionOptions: sdkNextRouteFactory.ClusterSolutionOptions{
				Depth: 100,
				Speed: 10,
			},
			FilterAreaOptions: sdkNextRouteFactory.FilterAreaOptions{
				MaximumWidth:  math.MaxFloat64,
				MaximumHeight: math.MaxFloat64,
				MaximumRadius: math.MaxFloat64,
			},
		},
		factory.NewDefaultModelFactory(),
	)
```

### DIFF
--- a/nextroute/factory/construction.go
+++ b/nextroute/factory/construction.go
@@ -18,15 +18,38 @@ type ClusterSolutionOptions struct {
 // FilterAreaOptions configure how the [NewGreedySolution] function builds [sdkNextRoute.Solution]. It limits the area
 // one vehicle can cover during construction. This limit is only applied during the construction of the solution.
 type FilterAreaOptions struct {
-	MaximumWidth  float64 `json:"maximum_width" usage:"maximum width of the area in meters" default:"100000" minimum:"0"`
-	MaximumHeight float64 `json:"maximum_height" usage:"maximum height of the area in meters" default:"100000" minimum:"0"`
-	MaximumRadius float64 `json:"maximum_radius" usage:"maximum radius of the area in meters" default:"100000" minimum:"0"`
+	MaximumSide float64 `json:"maximum_side" usage:"maximum side of the square area in meters" default:"100000" minimum:"0"`
 }
 
 // GreedySolutionOptions configure how the [NewGreedySolution] function builds [sdkNextRoute.Solution].
 type GreedySolutionOptions struct {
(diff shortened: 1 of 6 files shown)


nextroute/factory/construction.go
```diff
@@ -0,0 +1,218 @@
...
+
+	"github.com/nextmv-io/sdk/connect"
+	sdkNextRoute "github.com/nextmv-io/sdk/nextroute"
+	"github.com/nextmv-io/sdk/nextroute/common"
+	"github.com/nextmv-io/sdk/nextroute/schema"
+)
+
+// ClusterSolutionOptions configure how the [NewGreedySolution] function builds [sdkNextRoute.Solution].
+type ClusterSolutionOptions struct {
+	Depth int     `json:"depth" usage:"maximum failed tries to add a cluster to a vehicle" default:"10" minimum:"0"`
```
nilsbeck, line 14: should we call it something different that goes more with the description? like `MaxRetries` or something like that?

nextroute/factory/construction.go
```diff
@@ -0,0 +1,218 @@
...
+type StopCluster interface {
+	// Stops returns the stops in the stop cluster.
+	Stops() []schema.Stop
+	// Centroid returns the centroid of the stop cluster.
+	Centroid() schema.Location
+}
+
+// NewStopCluster returns a new stop cluster for the given stops.
+func NewStopCluster(
+	stops []schema.Stop) StopCluster {
```
nilsbeck, line 83: formatting is off

nilsbeck (approved):
I approve, since we discussed in slack the Intersection interface will be removed. If we want to change the `deep` field to something more self explanatory is up to you.

nextroute/factory/construction.go
davidrijsman: we can rename it but it is not a retry...

nilsbeck: yeah, fair enough. But you know what I mean.

PR #394, opened 2023-11-02
Adding custom duration that serializes nicely

larsbeck:
# Description

This adds a custom type `types.Duration` which the flag parsing library knows (we register it). Therefore when we use this type in options that are passed to the runner, this type's `MarshalJSON()` will be called and render a duration in a nice format rather than nanoseconds.


types/duration.go
```diff
@@ -0,0 +1,15 @@
...
+
+import "time"
+
+// Duration is a wrapper around time.Duration that implements json.Marshaler.
+type Duration time.Duration
+
+// MarshalJSON implements json.Marshaler. It calls time.Duration.String() to
+// serialize the duration.
+func (d Duration) MarshalJSON() ([]byte, error) {
+	return []byte(`"` + time.Duration(d).String() + `"`), nil
```
sebastian-quintero: Why the extra `"`?

sebastian-quintero (approved):

sebastian-quintero:
We might need to change the `shift-scheduling` template with this PR to reflect the use of the new type?

types/duration.go
larsbeck: Good catch. Removed

sebastian-quintero:
nextroute moved over to another repo.

PR #397, opened 2023-11-06
feature/doc-709-investigate-shift-scheduling-sample-input-in-docs

nilsbeck:


davidyeiser:
Is there a specific reason why the required shifts are narrow in range compared to available shifts? (See screenshot below)

Screen Shot 2023-11-06 at 4 44 45 PM

nilsbeck:
> Is there a specific reason why the required shifts are narrow in range compared to available shifts? (See screenshot below)
> 
> Screen Shot 2023-11-06 at 4 44 45 PM
> 

Technically no. From the customer for which we made this model we know that they had small time slots with their respective demand for drivers. While drivers usually give a wide range of availability they are basically available for the morning and or evening and don't want to work for only an hour.
And the smaller time slots are probably due to short peaks over lunch etc that do not span larger portions of the day (I assume). 

But the model could handle wider time slots.

davidyeiser:
Got it — Could we edit those required slots to either be longer or add more for the second block of shifts that overlap? (The ones from 9am to 12pm in the visual above.)

I'm mainly thinking from an initial experience with the app. It would be a good experience if it was very simple and something the user could parse visually and start to form a mental model of how it's working. That is I see shifts that need to be filled. I see a bunch of worker available shifts and know I have the capacity. I then see the result where everything is filled and orderly.

For example, right now we don't have enough workers to fill slot number 5. And then the result has a shift that goes way beyond the required worker bounds and looks kind of odd.

nilsbeck:
> Got it — Could we edit those required slots to either be longer or add more for the second block of shifts that overlap? (The ones from 9am to 12pm in the visual above.)
> 
> I'm mainly thinking from an initial experience with the app. It would be a good experience if it was very simple and something the user could parse visually and start to form a mental model of how it's working. That is I see shifts that need to be filled. I see a bunch of worker available shifts and know I have the capacity. I then see the result where everything is filled and orderly.
> 
> For example, right now we don't have enough workers to fill slot number 5. And then the result has a shift that goes way beyond the required worker bounds and looks kind of odd.

That the shift goes beyond the current required worker bound is due to the model. The reason is that the model optimizes for filling required worker slots but does _not_ optimize for short shift lengths of the workers.

davidyeiser:
Okay, that makes sense, thanks.

So I think to close this one out, let's make the required worker shifts longer so the solution doesn't have an extreme difference in shift lengths. See the demo to get an idea of what I'm talking about: https://cloud.stg.nxmv.xyz/marketplace/app/nextmv/shift.scheduling/demo/general

nmisek (approved):

PR #400, opened 2023-11-14
Extended cluster constraint and objective to exclude first and last stop

davidrijsman:
Added the option to configure the first and last stop of a vehicle to be considered in calculating the centroid. The default is to exclude them, this is a change of the current implementation and considered a bug fix.

### DIFF
--- a/nextroute/model_cluster.go
+++ b/nextroute/model_cluster.go
@@ -13,10 +13,30 @@ type Cluster interface {
 	ConstraintStopDataUpdater
 	ModelConstraint
 	ModelObjective
+
+	// IncludeFirst returns whether the first of the vehicle is included in the
+	// centroid calculation. The centroid is used to determine the distance
+	// between a new stop and the cluster.
+	IncludeFirst() bool
+	// IncludeLast returns whether the lst stop of the vehicle is included in
+	// the centroid calculation. The centroid is used to determine the distance
+	//between a new stop and the cluster.
+	IncludeLast() bool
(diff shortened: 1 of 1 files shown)


nextroute/model_cluster.go
```diff
@@ -13,10 +13,30 @@ type Cluster interface {
 	ConstraintStopDataUpdater
 	ModelConstraint
 	ModelObjective
+
+	// IncludeFirst returns whether the first of the vehicle is included in the
+	// centroid calculation. The centroid is used to determine the distance
+	// between a new stop and the cluster.
```
larsbeck:
```suggestion
	// IncludeFirst returns whether the first stop of the vehicle is included in the
	// centroid calculation. The centroid is used to determine the distance
	// between a new stop and the cluster.
```

nextroute/model_cluster.go
```diff
@@ -13,10 +13,30 @@ type Cluster interface {
...
+	IncludeFirst() bool
+	// IncludeLast returns whether the lst stop of the vehicle is included in
+	// the centroid calculation. The centroid is used to determine the distance
+	// between a new stop and the cluster.
```
larsbeck:
```suggestion
	// IncludeLast returns whether the last stop of the vehicle is included in
	// the centroid calculation. The centroid is used to determine the distance
	// between a new stop and the cluster.
```

larsbeck (approved):

PR #405, opened 2023-11-20
feature/eng-4203-custom-measure-for-looking-up-cost-in-a-unique-matrix

nilsbeck:
# Description

This PR adds a new Unique measure to our measures. It takes a unique matrix of costs and a slice of stop references which maps a stop by index (index from the input) to an index in the unique matrix.
The measure does not create the unique matrix etc itself and expects them to be unique. Reason is that this way a unique set of points can be requested from e.g. and OSRM server, and the result can directly be used with this type of measure.

* `unique.go`
* `unique_test.go`

### DIFF
--- a/measure/unique.go
+++ b/measure/unique.go
@@ -1,0 +1,37 @@
+package measure
+
+import "encoding/json"
+
+// Unique returns a ByIndex that uses a reference slice to map the indices of a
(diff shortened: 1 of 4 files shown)


merschformann (approved):
Nice, thanks! :hugs:

PR #407, opened 2023-11-27
feature/eng-4224-deprecate-legacy-sdk-stuff

nilsbeck:
# Description

Declares functions and packages as `deprecated` (see list below) and refactors some functions by moving to them to a more appropriate place. To avoid a breaking change, type aliases are introduced in those cases.
Because Go is unable to handle well all the `redirects` needed when `type aliasing` functions some functions, especially the matrix clients like `OSRM`, it was necessary to duplicate some code. I kept the duplicate code to a minimum. However, this should not mark a big difference since the idea is that this code will go away with the next major release.

I also tested that the new clients are functional in an external go project to make sure the clients in `measure/osrm` etc work as expected - I tested it by using the matrix-measure template with the new package.

Due to the deprecation VS Code and others will now mark packages and functions with strike-through (see image below).
Packages and their functions that are marked as deprecated:
- route
- route/osrm
- route/goroutingkit
- route/here
- route/google
- dataframe
- store

Example of strike-through:
image


common/alias.go
sebastian-quintero, line 1: Do we really need a `common` package at the top level of `sdk`? Go advises against having packages such as `common`, `utils`, etc... Aren't these things _just_ used inside `nextroute`? Shouldn't we keep them there somehow? Many moons ago I created this PR: https://github.com/nextmv-io/sdk/pull/248. Maybe you can read the reasoning there and take what you agree with, if anything at all.

dataframe/aggregation.go
```diff
@@ -4,30 +4,46 @@ import "fmt"
 
 // Aggregation defines how to aggregate rows of a group of rows in a Groups
 // instance.
+//
+// Deprecated: This package is deprecated and will be removed in a future.
```
sebastian-quintero:
```suggestion
// Deprecated: This package is deprecated and will be removed in the next major release.
```

Two opinions here:
- I think being clear about when it is going out of commission is better.
- If you don't like that for some reason, I would change `a future` for `the future`. You can check me with a native speaker but I think `the` is more appropriate here.


I'm making this comment just once to avoid polluting the PR with the same comment over and over again.

model/model.go
```diff
@@ -14,6 +14,9 @@ const (
 	MinInt = (1 << bits.UintSize) / -2
 )
 
+// Domain is deprecated.nextroute engine. It is used with the router engine and store
```
sebastian-quintero: I think these comments got messed up probably with a replace or something. Do you mean to add the `Deprecated: ` note here?

sebastian-quintero (approved):
I left some comments that are not necessarily blocking, so approving. Would be good if you can address them. Thanks!

common/alias.go
nilsbeck: This suggestion came from @davidrijsman And I agree that it makes sense since those are just very generic functions that work as helpers. currently, we may use them only in nextroute but that may change now that they are not part of nextroute anymore.

dataframe/aggregation.go
nilsbeck: made the change to use "next major release"

model/model.go
nilsbeck: cleaned them

PR #408, opened 2023-11-27
ENG-4254 Add the Pyomo template and Ruff formatter

sebastian-quintero:
# Description

- Adds the `pyomo` template.
- Changes the `black` and `flake8` formatter for `ruff`.
- Changes some linting issues on other Python files.


.github/workflows/python-lint.yml
merschformann, line 1: We should make sure,

1. that we announce this change for all of us to adapt their local setup to
2. align platform as well
3. reformat all python files to make sure they are in line with the new formatter (and options)

templates/pyomo/main.py
```diff
@@ -0,0 +1,158 @@
...
+    solution = solve(input_data, args.duration)
+    write_output(args.output, solution)
+
+
+def solve(input_data: dict[str, Any], duration: int) -> dict[str, Any]:
+    """Solves the given problem and returns the solution."""
+
+    # Creates the model.
+    model = pyo.ConcreteModel()
+    provider = "glpk"
```
cowanwalls: We might want to be more explicit about the fact we're defining a provider here, either in a comment in this file or in the template README. I could potentially see this being unintuitive to debug for CBC users depending on what error message they'd see when running this or if it simply would not behave as they'd expect.

cowanwalls (commented):
Looks solid to me but it's been quite a while since I've done any actual work in Python. I also don't want to approve before @merschformann's questions about the change in Python linter used are confirmed.

@sebastian-quintero if you ping me after those have been responded to I can come back in and approve.

templates/pyomo/main.py
sebastian-quintero: @cowanwalls Thanks. I changed the comment in the code, the error message and the README to make it more explicit that there are supported providers.

sebastian-quintero:
@merschformann @cowanwalls I processed your comments and made all the appropriate changes that both of you suggested.

.github/workflows/python-lint.yml
sebastian-quintero: Took care of everything 😄 .

cowanwalls (approved):

cowanwalls (approved):

PR #412, opened 2023-12-05
Allow configuration of operators

larsbeck:
# Description

This PR exposes more ALNS related constructs and introduces methods that allow users to add operators. On the `ParallelSolver` it adds capabilities to set a `NewSolverFactory` and `NewSolveOptionsFactory` which both can be used to influence which subsolvers are used by the `ParallelSolver` in each cycle.


alns/solve_options.go
```diff
@@ -0,0 +1,18 @@
...
+	return SolveOptions{
+		Iterations:         -1,
+		SolutionBufferSize: 100,
+	}
+}
+
+// SolveOptions contains the options for the solve process.
+type SolveOptions struct {
+	Iterations         int `json:"iterations" default:"-1"`
+	SolutionBufferSize int `json:"solution_buffer_size" default:"100"`
```
davidrijsman: What is the buffer size doing again, we should document now we are exposing it

davidrijsman (approved):

alns/solve_options.go
larsbeck: It configures the length of the solution channel, but instead of describing it, I decided to remove and configure it internally as I don't see that this needs to be configurable at this time. If we find out otherwise, we can re-introduce the field.

PR #416, opened 2024-01-08
feature/eng-4404-add-new-observer-interfaces-to-observe-initial-stops-better

nilsbeck:
# Description

Adds new observer interfaces.

### DIFF
--- a/nextroute/solution_observer.go
+++ b/nextroute/solution_observer.go
@@ -19,7 +19,7 @@ type SolutionObserver interface {
 		violation CheckedAt,
 	)
 
-	// OnCheckedConstraint is called when a constraint has been checked.
+	// OnSolutionConstraintChecked is called when a constraint has been checked.
 	OnSolutionConstraintChecked(
 		constraint ModelConstraint,
 		feasible bool,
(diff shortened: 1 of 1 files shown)


nextroute/solution_observer.go
```diff
@@ -70,6 +70,56 @@ type SolutionUnPlanObserver interface {
...
+	// been estimated.
+	OnEstimatedIsViolatedWithStop(
+		constraint ModelConstraint,
+		isViolated bool,
+		planPositionsHint StopPositionsHint,
+		stop ModelStop,
+	)
+}
+
+// OnConstraintCheckObserver is an interface to observe the constraint check.
```
davidrijsman: Why introduce a new API, this will be a breaking change anyway

nilsbeck: I don't think it is breaking. It was the intention to not make a breaking change.
I think so because it is a new interface unrelated old one. In other words, you still comply to older interfaces the same way but can add new observers by complying new interfaces.

davidrijsman: You are right, this is not brraking

davidrijsman (approved):

PR #417, opened 2024-01-16
Add support for pre-recorded output

dirkschumacher:
This adds a special key that can be used to allow pre-recorded output in the runner.

### DIFF
--- a/run/validate/json.go
+++ b/run/validate/json.go
@@ -15,6 +15,10 @@ import (
 	"github.com/xeipuuv/gojsonschema"
 )
 
+// RecordedOutputKey is the key used to store the pre-recored output in the
+// input itself. It can be used to handle pre-computed runs.
+const RecordedOutputKey = "__recorded_output"
+
 // JSON creates a JSON validator. If nil is passed as schema, the validator will
 // try to read schema.json in the current directory. If that file does not
 // exist, no validation will be performed.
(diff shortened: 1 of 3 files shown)


larsbeck (approved):

merschformann (approved):

dirkschumacher:
@merschformann @larsbeck any ideas on the name? `$recorded_output`, `_recorded_output`, `$pre_recorded_output`, `$output`. etc. etc.

larsbeck:
> @merschformann @larsbeck any ideas on the name? `$recorded_output`, `_recorded_output`, `$pre_recorded_output`, `$output`. etc. etc.

`__recorded_output` is good for me.

merschformann:
I thought the reserved field was `__recorded_output`. I don't have a strong opinion on the name. Is there any kind of convention around JSON? The leading double `_` may be easy to get wrong, but also not a real issue. Coming from Python one may take it as a _really_ private field. :relaxed:

dirkschumacher:
Yeah I am unsure. My intent was to clearly distinguish it from any user defined keys. `$recorded_output` might also be a good option.

merschformann:
I prefer the `_`, but that may just be my Python side. The `$` already has a different meaning in JSON context, right? I would avoid it, if that is the case. Doesn't it refer to the root or something?

sebastian-quintero (approved):
LGTM

PR #428, opened 2024-02-07
Feature/add-cost-flow-app

nilsbeck:
# Description

This PR adds a cost-flow app to the templates.


sebastian-quintero (approved):
LGTM, would add to community-apps too.

nilsbeck:
> LGTM, would add to community-apps too.

that was the plan - once this gets through. do not edit in 2 places all the time :)

templates/cost-flow-ortools/main.py
```diff
@@ -0,0 +1,281 @@
...
+from typing import Any
+
+from ortools.graph.python import min_cost_flow
+
+STATUS = {
+    min_cost_flow.SimpleMinCostFlow.OPTIMAL: "optimal",
+    min_cost_flow.SimpleMinCostFlow.BAD_COST_RANGE: "bad cost range",
+    min_cost_flow.SimpleMinCostFlow.BAD_RESULT: "bad result",
+    min_cost_flow.SimpleMinCostFlow.FEASIBLE: "feasible",
+    min_cost_flow.SimpleMinCostFlow.INFEASIBLE: "infeasible",
```
merschformann, line 18: Is it actually possible to have an infeasible result or any other result that has no solution values? If so, it would be great if we would test for it so that the app does not crash in those situations. We had/have this with some mip templates.

merschformann: One thing I found: if I add a required skill that is not available, I get a segfault. It would be good, if we could catch such situations.

merschformann (commented):
Looks great already. Just a minor ask to improve stability a bit more, if possible. I don't want to be blocking for this, if this is out-of-scope though. :blush:

merschformann (approved):

merschformann (approved):

PR #431, opened 2024-03-05
Allows changing working dir for test execution

merschformann:
# Description

Allows changing the working directory for the command being executed in the test. Input file paths will be adapted automatically.

## Changes

- Allows custom working directory in golden file testing

### DIFF
--- a/golden/config.go
+++ b/golden/config.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


golden/config.go
```diff
@@ -174,20 +178,31 @@ func (config Config) entrypoint(inputPath string) (*exec.Cmd, string, error) {
...
+
+	// When not handling I/O with stdin/stdout, use temporary files
 	if !config.UseStdIn {
 		inputFlag := "-runner.input.path"
 		if isCustom {
 			inputFlag = config.ExecutionConfig.InputFlag
 		}
 		args = append(args, inputFlag, inputPath)
 	}
-
```
sebastian-quintero: Matter of taste, I like a newline after a control block finishes (for, if, etc)

sebastian-quintero (approved):

golden/config.go
merschformann: The idea here was to group the different steps nicely. But yeah, maybe a bit aggressive.

PR #433, opened 2024-03-06
Switches to independent release workflow

merschformann:
# Description

As we are moving away from the plugin based distribution, we need to release SDK independently.

## Changes

- Updated _independent_ release workflow


sebastian-quintero (approved):
Lgtm

PR #436, opened 2024-03-14
More cleanup before v2.0.0

sebastian-quintero:
# Description

More cleanup:

- Updated Go workflows to use Go 1.22
- Removed Python workflows
- Updated linting file for Go 1.22. Removed packages that no longer exist.
- Updated `go.mod` to use Go 1.21.
- Fixed linting exceptions.
- Removed version logic to just reference the version of the project. The override version used to exist to jump hoops with plugins, it is not needed anymore.


version.go
```diff
@@ -5,39 +5,9 @@ package sdk
...
-		return dep.Version
-	}
-
-	// If this happens, we're running in a module in which sdk is not a
-	// dependency. In this case, we expect the NEXTMV_SDK_OVERRIDE_VERSION to be
-	// set. Thus, this is unexpected. So, return a string that helps us find the
-	// way back here.
-	return "no-overridden-version-found"
-}
+var VERSION string
```
merschformann: Do we really want to drop this? This still works fine and is more accurate. We only need to shuffle it around to fallback to `VERSION` instead of `"no-overridden-version-found"`.

merschformann (approved):

PR #443, opened 2024-03-19
Removes go setup step for lint action

merschformann:


merschformann:
This was just a test. We will keep the explicit go setup for now. Closing

PR #444, opened 2024-03-19
nilsbeck/fix-test-linting

nilsbeck:
# Description

Fix test linting.


.golangci.yml
```diff
@@ -199,3 +199,9 @@ issues:
         - gocyclo
         - nestif
         - gosec
+    # The test is only used internally and doesn't need
+    # to meet the same standards as the rest of the codebase for now.
+    - path: run/tests/http/async/main\.go
+      linters:
+        - gosec
+      text: G114
```
merschformann: Can you add the newline? :pleading_face:

run/tests/http/async/main.go
```diff
@@ -18,6 +18,7 @@ func main() {
 	go func() {
 		handler := http.HandlerFunc(callback)
 		http.Handle("/callback", handler)
+		//nolint:gosec
```
merschformann: Is this necessary, if we configure it in the yml?

nilsbeck: normally not, but it is not picking it up from the config. and I don't know why. since this is not an important test, I just want to go on here.

.golangci.yml
nilsbeck: sure...

merschformann (approved):

PR #447, opened 2024-03-29
Feature/eng 4608 add integration tests to the platform repo for the platform

nilsbeck:
# Description

This PR adds functionality to run bash tests in chained way. This is particularly useful when different tests depend on each other. Currently the tests are running in a sequence one after another but the framework has the ground work to run tests in parallel and await the test result before continuing.


merschformann (approved):
:tada:

PR #448, opened 2024-04-02
nilsbeck/run dag in parallel

nilsbeck:
# Description

This makes the DAG bash test run in parallel where the DAG allows for it. It also adds a WaitBefore field to the bash config to allow for a waiting time (in seconds) before the bash script is executed. This is useful if you have to wait for rate limits. If the value is set t 0 no delay is applied.

### DIFF
--- a/golden/config.go
+++ b/golden/config.go
@@ -79,8 +79,7 @@ type BashConfig struct {
 	// executed.
 	WorkingDir string
 	// WaitBefore adds a delay before running the bash script to handle rate limits.
-	// The delay is in seconds, if the value is set to 0 no delay will be added.
-	WaitBefore int
+	WaitBefore time.Duration
 }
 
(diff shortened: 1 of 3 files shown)


golden/config.go
```diff
@@ -78,6 +78,8 @@ type BashConfig struct {
 	// WorkingDir is the directory where the bash script(s) will be
 	// executed.
 	WorkingDir string
+	// WaitBefore adds a delay before running the bash script to handle rate limits.
```
merschformann:
```suggestion
	// WaitBefore adds a delay before running the bash script. This is useful
	// when throttling is needed, e.g., when dealing with rate limiting.
```

Just to suggest that there may be other uses for this in the future, as this is a shared package. :relaxed:

merschformann (approved):
:tada:

PR #451, opened 2024-04-09
nilsbeck/add-WithApproachCurb-to-TableOptions

nilsbeck:
# Description

This PR adds a new `TableOption` to enforce curb approaches when generating distance and duration matrices.

### DIFF
--- a/measure/osrm/client.go
+++ b/measure/osrm/client.go
@@ -516,8 +516,6 @@ func ParallelRuns(runs int) TableOptions {
 	}
 }
 
-//
-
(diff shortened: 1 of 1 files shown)


measure/osrm/client.go
```diff
@@ -498,6 +516,8 @@ func ParallelRuns(runs int) TableOptions {
 	}
 }
 
+//
+
```
merschformann:
```suggestion
```

merschformann (approved):

PR #453, opened 2024-05-20
Add a SleepBefore config for a DAG test case

sebastian-quintero:
# Description

Adds a new configuration for a `DagTestCase`: the `DagTestCaseConfig`. That configuration is introduced so that we may sleep before executing that step in the DAG. As a consequence, the old `Config` was renamed to `BashConfig`, as it is more appropriate.

### DIFF
--- a/golden/dag.go
+++ b/golden/dag.go
@@ -4,14 +4,28 @@ import (
 	"fmt"
 	"sync"
 	"testing"
+	"time"
 )
 
 // DagTestCase represents a test case in a directed acyclic graph (DAG) test.
 type DagTestCase struct {
-	Name   string
-	Needs  []string
-	Config *BashConfig
-	Path   string
+	// Bash configuration for this test case in the DAG.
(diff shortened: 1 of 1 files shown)


golden/dag.go
```diff
@@ -4,14 +4,28 @@ import (
...
 // DagTestCase represents a test case in a directed acyclic graph (DAG) test.
 type DagTestCase struct {
-	Name   string
-	Needs  []string
-	Config *BashConfig
-	Path   string
+	// Bash configuration for this test case in the DAG.
+	BashConfig *BashConfig
+	// Configuration for this DAG test case.
+	Config *DagTestCaseConfig
```
sebastian-quintero, line 15: These are the only actual changes. The rest of the properties change because I added godocs.

merschformann (approved):

sebastian-quintero:
A different functionality already achieves what I intended. Sorry.

PR #455, opened 2024-05-21
Enhanced OSRM Client with Exclude Parameter Support

merschformann:
# Description
This update introduces the ability to specify exclude parameters in OSRM table requests.

# Changes
- Added support for `exclude` parameter in table requests.
- Adds default of requesting distance AND duration, if nothing is requested explicitly


nilsbeck (approved):
Olé! 👏

PR #461, opened 2024-11-03
Fix typos

cnpryer:
These corrections are only in comments, but I ran [typos-cli](https://github.com/crate-ci/typos) on the repository because of _`Tresholds`_ ("thresholds").

https://github.com/nextmv-io/sdk/blob/9fe2ea13d7097c873057a36dba00372b72606c94/golden/config.go#L115-L132

I figured since that'd be a breaking change I wouldn't include it in this PR, but at least I'd point it out.

### DIFF
--- a/measure/haversine.go
+++ b/measure/haversine.go
@@ -27,7 +27,7 @@ func (h haversineByPoint) Cost(p1, p2 Point) float64 {
 	x := 2.0 * radius * math.Atan2(math.Sqrt(a), math.Sqrt(1.0-a))
 
 	// on different architectures x is slightly different. we don't need perfect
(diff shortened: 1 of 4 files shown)


merschformann (approved):
Sorry letting this one hang here for so long.
Thank you for the work! :hugs:

cnpryer:
No problem! Would you be interested in a PR to fix _Tresholds_?

merschformann:
> No problem! Would you be interested in a PR to fix _Tresholds_?

Yeah, it would be nice to fix it. Something that bugs me as well, now that you pointed it out. :blush: 

It being _breaking_ shouldn't be a problem. We only use it internally and upgrading it is a deliberate choice (`go get ...`).

PR #467, opened 2025-06-17
Rename `Tresholds` to `Thresholds`

cnpryer:
Following up on https://github.com/nextmv-io/sdk/pull/461#issuecomment-2532226121.

This fixes a typo. Note that the typo is related to the current `type` `Tresholds` -- which would technically be a breaking change. 

Sorry for taking so long! Lost track of this one. 

Let me know if you'd like this PR to include the appropriate version bump.

### DIFF
--- a/.github/workflows/go-lint.yml
+++ b/.github/workflows/go-lint.yml
@@ -1,5 +1,8 @@
 name: go lint
-on: [push]
+on:
+  push:
+  pull_request:
+    types: [opened, synchronize, reopened]
 jobs:
   sdk-go-lint:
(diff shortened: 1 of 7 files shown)


merschformann (approved):
We should do it, even if it is a breaking change. It's a simple fix and we don't want to carry the typo along. :blush:

merschformann:
@cnpryer : I hijacked your branch a bit to expand the events on which our blocking workflows run here. Otherwise, it makes it quite difficult to get the required checks out of the way to be able to merge. :blush:

